PR #111, opened 2016-09-24
Improve autocomplete

### LINKED ISSUE #58: Improve support to tab completion
Tab completion is extremely helpful in command line environment I think nash could improve this feature.

tiago4orion:
Auto complete is not paleo...

Trying to fix #58 

Nash now supports the autocompletes below:

```sh
# absolute path names
λ> /pr<TAB>
λ> /proc

# local files
λ> ls c<TAB>
ls cli.go                ls completer.go          ls completer_test.go 

# system wide PATH binaries
λ> gre<TAB>
gregorio          grep              grep-changelog    gresource
```

*Still buggy for system-wide binaries when there's lots of choices... *

@patito What do you think now? Can you test?

### DIFF
--- a/cmd/nash/completer.go
+++ b/cmd/nash/completer.go
@@ -32,7 +32,7 @@ func (c *Completer) Do(line []rune, pos int) (newLine [][]rune, offset int) {
 	completeStr := line
 
 	for i := pos - 1; i >= 0; i-- {
-		if line[i] == ' ' || i == 0 {
+		if line[i] == ' ' {
 			completeStr = completeStr[i+1 : pos]
 			local = true
 			break
--- a/cmd/nash/completer.go
+++ b/cmd/nash/completer.go
@@ -21,12 +21,35 @@ func NewCompleter(p ...readline.PrefixCompleterInterface) *Completer {
 }
 
 func (c *Completer) Do(line []rune, pos int) (newLine [][]rune, offset int) {
+	var local bool
+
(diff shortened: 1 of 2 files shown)


tiago4orion:
I tried to fix the autocomplete issue in several ways.. with no success, The bug apparently is on the readline third party lib. I forked the readline and modified the examples to add several choices in the autocomplete-mode and the bug shows there too.. I tried to fix there in several ways, with no success too. Jesus, never mind..

Until we find a better alternative to autocomplete, I'll leave this way. I'm seriously thinking about the idea of drop autocomplete and readline entirely and hack some better alternative.

tiago4orion:
I'm thinking in remove the autocomplete of nash and create an interface to hook up an autocomplete function defined in nash scripting. This way the autocomplete could be implemented as a third party tool. Lots of buggy code could be removed.

As an example, the project [fzf](https://github.com/junegunn/fzf) could be used as autocomplete framework, only requiring to declare a special function in nash syntax:

```sh
fn nash_autocomplete(line, pos) {
        path = "."
        IFS = ()
        # set path accordinly to line
        choice <= find $path | fzf -q $line
        return $choice
}
```
Then, each time the user press <TAB> in the shell this function will be invoked and the result (if any) will be "autocompleted" in the command prompt.

lborguetti:
Yep dude! autocomplete isn't paleo ... :cry:

Issue #114, opened 2016-10-01
Unexpected token parsing statement '='


katcipis:
The line that generates this is:

```
volumes <= docker volume ls -qf dangling=true
```

I'm not sure if there is an error on this, the command on the right side of the statement is a valid shell command, should I quote it ?

Also I'm a little worried because an error on a alias aborted nash, If i place it as my default shell (as I intend to) if I make some mistake on an alias it will stop to start when I login on the machine. It would be good to know that one of the aliases  broke, but aborting seems a little extreme on this case. When executing other types of automation this seems great, but as the default shell that is loaded when I run nash it really seems extreme.

tiago4orion:
Currently, on nash you need to quote every argument that uses '=', but I'm not sure if it's the best behaviour..
The reason is to avoid the ambiguity below:

```sh
ls = "a"
```
The statement above must be interpreted as a variable assignment (LS EQUAL A) or as a command execution (CMD[ls] ARG[=] STRING[a]) ?

Currently the command is a variable assignment.

```sh
λ> ls = "a"
λ> echo $ls
a
λ> touch "=" a
λ> ls "=" a
'='  a
```

tiago4orion:
About the second problem related to nash aborting execution on first error found, as you know, this is required for nash scripts, but today nash works the same way (at least very close) when executed interactive and as script.

If you made some mistake on your init, you can start nash with "-noinit" parameter to start it fresh. What do you think? Solves the problem?

katcipis:
How does shell solve this ? I think it works on shell/bash.

About running with -noinit, it actually wont work. An error on that nature will make you unable to log in your machine depending on the distro, fixing it is not rocket science, but it is also not fun and pleasant :-).

I understand the beauty in letting it symmetric with how scripts are executed, but I'm a little torn with the aggressive behavior on failure when using as a interactive shell (and understanding a little better the bash behavior also :P)

tiago4orion:
bash/posix shells are bizarre:
```sh
[i4k@johncarter ~]$ ls = "a"
ls: cannot access '=': No such file or directory
ls: cannot access 'a': No such file or directory
[i4k@johncarter ~]$ ls="a"
[i4k@johncarter ~]$ echo $ls
a
[i4k@johncarter ~]$
```

PR #124, opened 2016-10-06
Hack to enable '=' in the middle of command arguments.

tiago4orion:
WIP: PoC to verify alternatives to syntax.

The problem is that '=' cannot be used in command arguments without quoting to avoid the ambiguity below (see issue #114 ):

    ls = "a"

It's a variable assignment? or a 'ls' listing on files '=' and "a"?

Current syntax is very straightforward, you must be explicit about what you're writing. Then, if you want execute listing on files '=' and 'a', you must quote the '='.

    ls "=" a
or
    ls "=" "a"

Otherwise, it's an assignment.

Ok, then?

This commit leaves the problem above as is, but enable '=' in other cases, like below:

    make collection=project1

or

    docker run -e VAR1= -e VAR2= something

But EQUAL is still *prohibited* in the cases below:

    1. make otheroption =something # (solvable)
    2. make otheroption = othervalue # (solvable) 
    3. make = something # (solvable, but only because assignment require quotes in the rhs)
    4. make = "something" # (impossible, this is an assignment by spec...)

1 & 2 are easy.
Solving 3 will make the parser code hard to follow or will require changes in the lexer to emit context to each token.

Maybe this confuses the syntax, and people could be more *surprise* of this behaviour than the enforcement of quoting '=' ... I don't know.

@katcipis @lborguetti @patito What do you guys think? It's a good direction?

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -466,7 +466,31 @@ func (p *Parser) getArgument(allowArg, allowConcat bool) (ast.Expr, error) {
 		arg = ast.NewStringExpr(firstToken.FileInfo, firstToken.Value(), true)
 	} else {
 		// Arg && Ident
-		arg = ast.NewStringExpr(firstToken.FileInfo, firstToken.Value(), false)
+		it = p.peek()
+
+		var argstr string
+
+		if it.Type() == token.Assign &&
+			// check if '=' is close to arg
+			firstToken.Column()+len(firstToken.Value()) == it.Column() {
+			assignIt := p.next()
+
+			it = p.peek()
+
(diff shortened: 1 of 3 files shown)


tiago4orion:
Addendum:

Currently, the commands below are valid ones:

```sh
./configure --prefix=/usr
go tool cover -html=c.out
```

The reason is that the ambiguity exists only when using '=' with token.Ident. In the cases above, "--prefix" and "-html" are tokens of type token.Arg.

katcipis:
For me the cases that are already covered are good enough :-)

tiago4orion:
OK.

PR #132, opened 2016-11-09
Only import local files if path is absolute

tiago4orion:
Workaround for the "import strategy issue".

/cc @katcipis


katcipis:
LGTM

PR #139, opened 2016-11-12
Remove IFS variable support.

### LINKED ISSUE #128: Proposals for new IFS implementation
It's very annoying that your command assignment depends on a external variable be set. It's error prone and can break in unexpected ways.

Because everyone can set a default value for IFS in their init, the scripts should always be careful and set a proper value before any command assignment. 

There are two ways I can think of to solve this problem. Remove IFS entirely or make it scoped.
## No IFS (No split by default)

_backward incompatible_

This way, every command assignment will always return a string and we'll need a built in function to split it explicitly into a list:

``` sh
val <= cat data.csv | grep -v "header" # $val store entire string
val <= split($val, ("\n")) # Split data into a list and store in $val
```
## Scoped IFS

_backward compatible_ 

Using IIFE (immediately-invoked function expression):

``` sh
val <= fn () {
        IFS = ("\n")
        ret <= cat data.csv | grep -v "header"
        return $ret
}()
```

It`s not supported yet, but easy to implement and works, but as this kind of assignment occurs very often, we need a shorter way...

Maybe a built in `split` command in the pipe:

``` sh
val <= cat data.csv | grep -v "header" | split "\n"
```

But it must be the last component of the pipeline...

Maybe improve the syntax to support scoped variables in statements:

``` sh
val <= (echo $path) path = $PATH
val <= (cat data.csv | grep -v "header") IFS = ("\n")
```

I cannot think of a better way to improve the syntax to support scoped variables...  Maybe someone could help.

@katcipis @lborguetti 
Ideas?

tiago4orion:
The `IFS` special variable was removed to increase script reliability. For now, if you want to split variable
content or command output into a list you can use the `split` builtin function.

    content = "some thing"
    list <= split($content, " ")
    echo $list[0] $list[1]

Closes #128 

Signed-off-by: Tiago <tiago.natel@neoway.com.br>

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,13 +78,12 @@ echo $fullpath
 The symbol '<=' redirects the stdout of the command or function invocation in the
 right-hand side to the variable name specified.
 
-If you want the command output splited into an array, use the IFS
-special variable to store the delimiter (like bash, but use a list
-instead of a string).
+If you want the command output splited into an array, then you'll need
+to store it in a temporary variable and then use the builtin `split` function.
 
 ```sh
-IFS = ("\n")
-files <= find .
+out <= find .
+files <= split($out, "\n")
 
 for f in $files {
         echo "File: " + $f
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -411,6 +411,10 @@ func (shell *Shell) setupBuiltin() error {
 	shell.builtins["append"] = appendfn
 	shell.Setvar("append", sh.NewFnObj(appendfn))
 
+	splitfn := NewSplitFn(shell)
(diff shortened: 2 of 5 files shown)


tiago4orion:
I forgot to update docs...

The following cases are addressed in this PR and need to be documented too:

Split by list of symbols:

```sh
separators = ("," ";" " " "\n")
list <= split($content, $separators)
```

Split by custom function:
```sh
# split functions must return "0" if content must be splited by `char`
# and any other value otherwise.
fn mySplit(char) {
        if $char == "," {
                return "0"
        } else if $char == "|" {
                return "0"
        }

        return "1"
}

values <= split($content, $mySplit)
```
}

PR #148, opened 2016-11-20
Add support for fncall in return

tiago4orion:
Add the last missing syntax construct for functions:

```sh
fn a() {
    return "1"
}

fn b() {
    return a()
}
```


katcipis:
Nice :-)

PR #152, opened 2016-12-28
Disable bindfn in non-interactive mode. 

### LINKED ISSUE #62: Fix OSX travis setup
@patito Can you help me?

Take a look into .travis.yml file.

Thanks!

### LINKED ISSUE #150: Disable binds made with 'bindfn' when running in non-interactive (script) mode
Running `nash` in non-interactive mode with binds enabled could be dangerous when running a script in different servers. `nash` was always guaranteed to run real commands instead of binds, but that will probably change in the future with #151.

### LINKED ISSUE #151: Make binds from `bindfn` have precedence over commands
This depends on #150.

`bindfn` should be used with the idea of overriding commands, so when you do `bindfn foo_function ls` the new `ls` execution should call `foo_function` instead of the command `ls`.

Right now `nash` will only execute a "binded" function if it doesn't find the equivalent command in the OS. Making it impossible to do what was previously cited.

vitorarins:
This PR got a little too long because In order to fix #62 I had to implement #151 first. Then as discussed in #150 I had to guarantee that `bindfn` would be disabled in non-interactive (script) mode before I could implement the binds precedence over commands.
Fixes #62 .
Closes #151 .
Closes #150 .

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -117,7 +117,7 @@ func (e *errStopWalking) StopWalking() bool { return true }
 func NewShell() (*Shell, error) {
 	shell := &Shell{
 		name:        "parent scope",
-		interactive: true,
+		interactive: false,
 		isFn:        false,
 		logf:        NewLog(logNS, false),
 		nashdPath:   nashdAutoDiscover(),
@@ -246,6 +246,18 @@ func (shell *Shell) SetDebug(d bool) {
 // SetInteractive enable/disable shell interactive mode
 func (shell *Shell) SetInteractive(i bool) {
 	shell.interactive = i
+
+	if i {
+		_ = shell.setupDefaultBindings()
+	}
+}
+
+func (shell *Shell) Interactive() bool {
+	if shell.parent != nil {
+		return shell.parent.Interactive()
+	}
+
+	return shell.interactive
 }
 
 func (shell *Shell) SetName(a string) {
@@ -411,7 +423,7 @@ func (shell *Shell) String() string {
 	return string(out.Bytes())
 }
 
-func (shell *Shell) setupBuiltin() error {
+func (shell *Shell) setupBuiltin() {
 	lenfn := NewLenFn(shell)
 	shell.builtins["len"] = lenfn
 	shell.Setvar("len", sh.NewFnObj(lenfn))
@@ -427,7 +439,9 @@ func (shell *Shell) setupBuiltin() error {
 	chdir := NewChdir(shell)
 	shell.builtins["chdir"] = chdir
 	shell.Setvar("chdir", sh.NewFnObj(chdir))
+}
 
+func (shell *Shell) setupDefaultBindings() error {
 	// only one builtin fn... no need for advanced machinery yet
 	err := shell.Exec(shell.name, `fn nash_builtin_cd(path) {
             if $path == "" {
@@ -455,7 +469,8 @@ func (shell *Shell) setup() error {
 		shell.Setvar("PROMPT", pobj)
 	}
 
-	return shell.setupBuiltin()
+	shell.setupBuiltin()
+	return err
 }
 
 func (shell *Shell) setupSignals() {
@@ -1201,7 +1216,7 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
 		shell.logf("Executing bind %s", cmdName)
 		shell.logf("%s bind to %s", cmdName, fn)
 
-		if !shell.interactive {
+		if !shell.Interactive() {
 			err = errors.NewEvalError(shell.filename,
 				c, "'%s' is a bind to '%s'. "+
 					"No binds allowed in non-interactive mode.",
@@ -2022,7 +2037,7 @@ execDump:
 }
 
 func (shell *Shell) executeBindFn(n *ast.BindFnNode) error {
-	if !shell.interactive {
+	if !shell.Interactive() {
 		return errors.NewEvalError(shell.filename,
 			n, "'bindfn' is not allowed in non-interactive mode.")
 	}
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -37,13 +37,14 @@ type (
 
 	// Shell is the core data structure.
 	Shell struct {
-		name      string
-		debug     bool
-		lambdas   uint
-		logf      LogFn
-		nashdPath string
-		isFn      bool
-		filename  string // current file being executed or imported
+		name        string
+		debug       bool
+		interactive bool
+		lambdas     uint
+		logf        LogFn
+		nashdPath   string
+		isFn        bool
+		filename    string // current file being executed or imported
 
 		interrupted bool
 		looping     bool
@@ -115,20 +116,21 @@ func (e *errStopWalking) StopWalking() bool { return true }
 // NewShell creates a new shell object
 func NewShell() (*Shell, error) {
 	shell := &Shell{
-		name:      "parent scope",
-		isFn:      false,
-		logf:      NewLog(logNS, false),
-		nashdPath: nashdAutoDiscover(),
-		stdout:    os.Stdout,
-		stderr:    os.Stderr,
-		stdin:     os.Stdin,
-		env:       make(Env),
-		vars:      make(Var),
-		fns:       make(Fns),
-		builtins:  make(Fns),
-		binds:     make(Fns),
-		Mutex:     &sync.Mutex{},
-		filename:  "<interactive>",
+		name:        "parent scope",
+		interactive: true,
+		isFn:        false,
+		logf:        NewLog(logNS, false),
+		nashdPath:   nashdAutoDiscover(),
+		stdout:      os.Stdout,
+		stderr:      os.Stderr,
+		stdin:       os.Stdin,
+		env:         make(Env),
+		vars:        make(Var),
+		fns:         make(Fns),
+		builtins:    make(Fns),
+		binds:       make(Fns),
+		Mutex:       &sync.Mutex{},
+		filename:    "<interactive>",
 	}
 
 	err := shell.setup()
@@ -239,6 +241,11 @@ func (shell *Shell) Reset() {
 func (shell *Shell) SetDebug(d bool) {
 	shell.debug = d
 	shell.logf = NewLog(logNS, d)
+}
+
+// SetInteractive enable/disable shell interactive mode
+func (shell *Shell) SetInteractive(i bool) {
+	shell.interactive = i
 }
 
 func (shell *Shell) SetName(a string) {
@@ -1194,6 +1201,15 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
 		shell.logf("Executing bind %s", cmdName)
 		shell.logf("%s bind to %s", cmdName, fn)
 
+		if !shell.interactive {
+			err = errors.NewEvalError(shell.filename,
+				c, "'%s' is a bind to '%s'. "+
+					"No binds allowed in non-interactive mode.",
+				cmdName,
+				fn.Name())
+			return nil, ignoreError, err
+		}
+
 		if len(c.Args()) > len(fn.ArgNames()) {
 			err = errors.NewEvalError(shell.filename,
 				c, "Too much arguments for"+
@@ -2006,6 +2022,11 @@ execDump:
 }
 
 func (shell *Shell) executeBindFn(n *ast.BindFnNode) error {
+	if !shell.interactive {
+		return errors.NewEvalError(shell.filename,
+			n, "'bindfn' is not allowed in non-interactive mode.")
+	}
+
 	if fn, ok := shell.GetFn(n.Name()); ok {
 		shell.Setbindfn(n.CmdName(), fn)
 	} else {
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -1190,6 +1190,29 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
 			c, "Empty command name...")
 	}
 
+	if fn, ok := shell.Getbindfn(cmdName); ok {
+		shell.logf("Executing bind %s", cmdName)
+		shell.logf("%s bind to %s", cmdName, fn)
+
+		if len(c.Args()) > len(fn.ArgNames()) {
+			err = errors.NewEvalError(shell.filename,
+				c, "Too much arguments for"+
+					" function '%s'. It expects %d args, but given %d. Arguments: %q",
+				fn.Name(),
+				len(fn.ArgNames()),
+				len(c.Args()), c.Args())
+			return nil, ignoreError, err
+		}
+
+		for i := 0 + len(c.Args()); i < len(fn.ArgNames()); i++ {
+			// fill missing args with empty string
+			// safe?
+			c.SetArgs(append(c.Args(), ast.NewStringExpr(token.NewFileInfo(0, 0), "", true)))
+		}
+
+		return fn, ignoreError, nil
+	}
+
 	cmd, err = NewCmd(cmdName)
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/sh/shell.go
```diff
@@ -241,6 +243,11 @@ func (shell *Shell) SetDebug(d bool) {
 	shell.logf = NewLog(logNS, d)
 }
 
+// SetInteractive enable/disable shell interactive mode
+func (shell *Shell) SetInteractive(i bool) {
```
katcipis: I'm not completely sure but I think it is more idiomatic to Go code to avoid Set/Get naming conventions. On this case even accessing a struct directly would be possible since you are allowing the data structure to be changed anyway. If there is any fear of in the future having to add some logic around the operation the name of the method could be just **Interactive** and you would call like:

```
a.Interactive(true)
```

It is one of the things that I really liked about Go naming convention, I'm fighting to drop this Get/Set pattern.

@tiago4orion reads much more Go code than I, he should know if this makes sense.

internal/sh/shell.go
```diff
@@ -1190,6 +1197,38 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
...
+				c, "Too much arguments for"+
+					" function '%s'. It expects %d args, but given %d. Arguments: %q",
+				fn.Name(),
+				len(fn.ArgNames()),
+				len(c.Args()), c.Args())
+			return nil, ignoreError, err
+		}
+
+		for i := 0 + len(c.Args()); i < len(fn.ArgNames()); i++ {
+			// fill missing args with empty string
```
katcipis: Not sure if I got this right, but the default behavior if I call a function with less args is to fill it in with empty strings ?

internal/sh/shell_test.go
```diff
@@ -911,24 +909,92 @@ func TestExecuteBindFn(t *testing.T) {
...
 		return
 	}
 
-	if strings.TrimSpace(string(out.Bytes())) != "override builtin cd" {
-		t.Errorf("Error: '%s' != 'override builtin cd'", strings.TrimSpace(string(out.Bytes())))
-		return
+	shell.SetInteractive(false)
+	shell.filename = "<non-interactive>"
+
+	expectedErr := "<non-interactive>:2:8: "+
```
katcipis: Here lies a interesting tradeoff, some people argue that sometimes it is necessary to inspect on tests the contents of an error message to validate that the right kind of error is generated. Well, in production code it would be a bad way to check for an error type, for me this is a hint that it is not a great idea for testing either, since it is good to test just as the user consumes the API.

Of course sometimes we mock stuff for example, creating the same kind of coupling/brittleness that this generates.

But what annoys me most on this approach is that it is actually worse than mocking. If you mock an interface and something changes you can get good compile time errors, this approach for testing will just give runtime errors, and this will happen when someone just improves an error message to be more clear, it is kind odd specially because the string representation of errors is supposed to be for debugging only (consumable by humans).

One alternative is to just check if an error is generated, this approach usually gives space for the test to pass without the right code implemented, but this usually is a bug on the test...if it exercises the right condition the error can happen only if you implement the error generating code (TDD approach). Usually I just do that.

If for some reason I'm feeling skeptic there is the approach where you can use sentinel errors, like io.EOF (or something like it). Or create sentinel types, like this:

https://golang.org/pkg/os/#LinkError

If the client does not need this kind of rich error handling, them I usually just check for the error and that is it. Of course there is tradeoffs involved, no easy answers here :-).

internal/sh/shell.go
vitorarins: It makes sense. I was going to follow the pattern of `SetDebug`, but I only realized now that `SetDebug` is a different case because it has logic on the operation like you said.

Will fix it right away :wink:

internal/sh/shell.go
vitorarins: Yes, you got it right. This code was here already, I just moved from place and decided not change logic. Maybe @tiago4orion can help with this one.

internal/sh/shell_test.go
vitorarins: I agree with you that we should not check for the whole content of the error. But I think the right error should be displayed because the user has to be given a good feedback about the error and how to fix it. So I liked the idea of sentinel errors/types, I just think we need more time to implement that.

internal/sh/shell.go
katcipis: Ow yeah, it is for the getter not the setter, forgot that :-).

It is an old personal opinion of mine that if you have Set/Get for something it does not make sense to make it private. It is like going for the illusion that stuff is encapsulated, but for all effects I can change the data structure.

There is a lot of objects in the http module where you just access stuff directly, no set/get, and I find it to me more simple and clean. But also it is not a huge deal.

Uniformity in Go code is one of the big pluses of Go, you should not take it as rules..but honestly don't caring to Go naming conventions does not seem like a good idea neither. These are very experienced people trying to come up with good naming conventions to make code easier to read, it is important ;-)

vitorarins: Sorry for the wrong choice of words, I generally like to follow the guidelines presented with the language. But a lot of stuff was designed by a lot of experienced people (Java, C++, generics, etc.). I am not underrating any of these works but what might work well for some people might not work well for others. So what I was trying to say is that I think the pattern is ok, but If we feel we should use something different and it is a consensus then we should go for it.

But namings aside, do you think I should export that variable?

internal/sh/shell.go
```diff
@@ -2006,6 +2022,11 @@ execDump:
 }
 
 func (shell *Shell) executeBindFn(n *ast.BindFnNode) error {
+	if !shell.interactive {
+		return errors.NewEvalError(shell.filename,
+			n, "'bindfn' is not allowed in non-interactive mode.")
```
vitorarins: @katcipis @tiago4orion I am not sure about this part of the code. This will break a lot of aliases scripts out there. It will also break [nashget](https://github.com/NeowayLabs/nashlib/blob/master/get.sh#L17).

One solution suggested by @tiago4orion was this:

> Export a variable like $NASHMODE with value "script" or "cli". And use this value for conditional imports in libraries.

```shell
 if $NASHMODE == "cli" {
   import "nashlib/aliases/all"
 }
```

Personally I think we can remove this and continue to accept `bindfn` on script mode, only the binds won't work. But this means that we can continue to create a bunch of new "commands" directly from the script, without the need for creating a new executable at `/usr/local/bin/something`.

internal/sh/shell.go
tiago4orion: I mostly agree with @katcipis arguments, the `http` package is a good example to follow but sometimes isn't only setting a property, involves some other logic, that then break the convention.

In case of `SetInteractive` maybe a method is the best option.

internal/sh/shell.go
```diff
@@ -115,20 +116,21 @@ func (e *errStopWalking) StopWalking() bool { return true }
...
-		stdin:     os.Stdin,
-		env:       make(Env),
-		vars:      make(Var),
-		fns:       make(Fns),
-		builtins:  make(Fns),
-		binds:     make(Fns),
-		Mutex:     &sync.Mutex{},
-		filename:  "<interactive>",
+		name:        "parent scope",
+		interactive: true,
```
tiago4orion: By default, interactive must be false. Setting the shell as interactive implies stdin and stdout properly set. If it's true by default, will break some tooling in Go.

internal/sh/shell.go
```diff
@@ -1190,6 +1197,38 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
 			c, "Empty command name...")
 	}
 
+	if fn, ok := shell.Getbindfn(cmdName); ok {
+		shell.logf("Executing bind %s", cmdName)
+		shell.logf("%s bind to %s", cmdName, fn)
+
+		if !shell.interactive {
```
tiago4orion: The right way to verify if the shell is in "interactive mode" must be asking the global interpreter (or the parent parent parent ... shell). For this, the acessor of the `interactive` property must be something like:

```go
func (sh *Shell) Interactive() bool {
    if sh.parent != nil {
        return sh.parent.Interactive()
    }

    // global interpreter (the grand grand ... father)
    return sh.interactive
}
```

tiago4orion: Don't forget that functions are sub-shells, then the source of truth is the parent of all subshells.

internal/sh/shell.go
tiago4orion: One more reason I think `binds` must work only when in cli.
What happens here is that bindfn create a fake command that points to a function but with arguments *optional*.

Binding the command name `cd` to function `chdir(path)` would enforce cd having an argument (always). Filling missing arguments with empty string allows the common `cd` behaviour as expected:

```sh
fn cd(path) {
    if $path == "" {
        path = $HOME
    }

    chdir($path)
}

# no arguments, must go to $HOME
$ cd
$ cd /
```

Bindfn is a helper for cli, then not so much strictly.

This is the *why* of this piece of code, but maybe not the best option. We can discuss it later, maybe another issue.

internal/sh/shell_test.go
tiago4orion: I dont think sentinel errors apply either.

Why errors are checked? Because as @vitorarins said, it must inform the right line/column and message. 

In the past I'd not checked the error string, but then @lborguetti found cases where line/column of the error was missing... Then I added this, and updated all tests to make sure line/column is always there to help...

internal/sh/shell_test.go
```diff
@@ -911,24 +909,92 @@ func TestExecuteBindFn(t *testing.T) {
...
+			t.Errorf("Error differs: Expected '%s' but got '%s'",
+				expectedErr, err.Error())
+		}
+	} else {
+		t.Errorf("An error was expected: '%s'. But did not occured.",
+			expectedErr)
+	}
+}
+
+func TestExecuteBindFn(t *testing.T) {
```
tiago4orion: This tests will fail after changes requested done. Requires a sh.SetInteractive(true).

internal/sh/shell.go
vitorarins: Ok. I got confused by the `filename` attribute which defaults to '<interactive>' and by the option `-i` described here: https://github.com/NeowayLabs/nash/blob/master/cmd/nash/main.go#L35 .

But I will change it right away.

internal/sh/shell.go
vitorarins: Right! Forgot that. Will fix it.

internal/sh/shell_test.go
vitorarins: I don't see a problem in making sure the line/column is there. But do we really need to validate **which** line and **which** column?

internal/sh/shell.go
vitorarins: Yes. Sorry about the noise.

internal/sh/shell_test.go
```diff
@@ -911,24 +909,92 @@ func TestExecuteBindFn(t *testing.T) {
...
 
 	if err != nil {
 		t.Error(err)
 		return
 	}
 
-	if strings.TrimSpace(string(out.Bytes())) != "override builtin cd" {
-		t.Errorf("Error: '%s' != 'override builtin cd'", strings.TrimSpace(string(out.Bytes())))
-		return
+	shell.SetInteractive(false)
```
vitorarins: Right. Will fix it.

internal/sh/shell_test.go
vitorarins: I am aware. Thank you for the detailed review 😉 .

internal/sh/shell_test.go
vitorarins: Anyway, like I said, I think this should be discussed as a different issue.

internal/sh/shell.go
tiago4orion: Yeah, the `nash` command is interactive by default (to be handy), but using the nash API (in Go) it's in script mode by default (to be handy also). ;)

When exec'ing `nash script.sh` it executes in non-interactive mode. But exec'ing `nash -i script.sh` it creates a shell session, executes the script.sh and prompt for new commands in the same shell (interactive).

vitorarins: Actually there is another reason for that. In the `NewShell` function we call `setup()` and the setup is executing: `bindfn nash_builtin_cd cd`. So if interactive mode is not true here, this will not work according to our rules.

tiago4orion: Ok, got it.

Interactive must be false when I simple make `NewShell()`, but default bindfns should not run.

What do you guys think of execute *default bindings* only when someone call SetInteractive(true) ?

internal/sh/shell.go
katcipis: Consistency wins on this case, if the rest of Nash code already uses Set functions, the function is definitely better. If this was the first instance of the problem, I would use direct access to the struct.

internal/sh/shell.go
katcipis: @tiago4orion definitely another issue :-). But thanks for showing the why, really makes sense on this case.

internal/sh/shell_test.go
katcipis: I understand your problem @tiago4orion, if line and collumn are this important to the correctness of the code I really would like to see them modeled on the domain logic, like:

https://golang.org/pkg/os/#LinkError

The LinkError struct has some fields with the details, and these could be tested directly. I understand that formatting properly can be pretty useful, but I had a lot of problems in the past with these kind of stuff. For example, an extra space can make your test break, and catching extra spaces on error messages is not trivial. It is tempting to go in this way, but for me it is the last of the last of the options, it opens the way to very brittle tests, in my opinion.

Bu, it is an interesting discussion for another time indeed, I have no problem with this being merged :-)

tiago4orion: @katcipis I fully agree with you, we need something like that.

We have a `NashError` struct, but it miss the line/col fields.
https://github.com/NeowayLabs/nash/blob/master/errors/error.go#L11

Line and column are'nt stored, but created by helper function. https://github.com/NeowayLabs/nash/blob/master/errors/error.go#L40
My fault, very bad indeed.

Adding this fields we can avoid making parsing of error message.

internal/sh/shell.go
tiago4orion: Do not apologize, your question was relevant. 
:)

internal/sh/shell.go
tiago4orion: @vitorarins Do you get the idea?

Something like:

```go
func (sh *Shell) SetInteractive(b bool) {
    sh.interactive = b

    if b {
        sh.setupBindings()
    }
}
```

internal/sh/shell_test.go
katcipis: Sounds good to me. If formatting is very very important you could also test the Error() method of the error type, but it will fall on the same problems discussed above. Perhaps it would be a little better because it could be tested on a very isolated manner.

internal/sh/shell.go
katcipis: If apologizing is required I will have to put it as a signature on my comments :-)

tiago4orion: hahaha

internal/sh/shell.go
vitorarins: I got the idea, I just don't think it would be good to add that much logic into a single `Set`. I will take a look at it and get back to you if I have another option. Maybe @katcipis can also share some opinions on this. 

This is going to take some time to review the test cases (a lot of them are going to fail).

tiago4orion:
@vitorarins Please, next time contribute with branches instead of fork. You have the rights to do so, much better to try your implementation. :)

vitorarins:
@tiago4orion I made some changes to `SetInteractive` like you suggested. But I still fail to see how can we allow the init file to be imported/executed with `bindfn`'s.

tiago4orion:
@vitorarins You'll need to arrange the `cmd/nash/main.go` to load the init only in interactive mode.

Scripts should not rely on things setup in the init file, because it's user configuration. Nash doesn't have system wide inits, maybe in the future if needed.

vitorarins:
@tiago4orion, I made the changes you suggested. Take a look when you can 😉 .

PR #154, opened 2017-01-06
Add shell.ExecFileArgs

katcipis:
Not sure if this is a great idea, but I was needing to call nash scripts from Go code, and my scripts take command line args. Checking out the current code I was unable a way to do so, but I found out code on the main package that did it and basically moved it. Probably not on the best way possible. Please advise :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 ./cmd/cnt/cnt
 /cmd/nash/nash
 /coverage.txt
+cmd/nashfmt/nashfmt
--- a/cmd/nash/main.go
+++ b/cmd/nash/main.go
@@ -41,9 +41,9 @@ func init() {
 }
 
 func main() {
-	var args  []string
+	var args []string
 	var shell *nash.Shell
-	var err   error
+	var err error
 
 	cliMode := false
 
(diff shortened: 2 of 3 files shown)


katcipis:
I think I may have missed some testing opportunity :-), if the idea is cool I can commit to write some tests.

vitorarins:
Looks like a good idea to me. But I am no expert. @katcipis maybe you can create an issue and give a use case for us, just to make it clear.

katcipis:
@vitorarins basically:

```
somescript.sh param1 param2
```

And from Go code I can call:

```
shell.ExecFileArgs("somescript.sh", "param1", "param2")
```

Right now I copied code from the main package to be able to do that,
[here](https://github.com/NeowayLabs/klb/blob/fixTests/tests/lib/nash/nash.go#L26).

tiago4orion:
> And current use of ExecFile without args would still work too, it seems better.

I prefer this way =) I just was dumb and forgot about script arguments :/

katcipis:
@tiago4orion shut up, you remember to implement the whole language right ? ;-)

vitorarins:
@katcipis great! LGTM.

katcipis:
@tiago4orion @vitorarins what about it now ?

tiago4orion:
LGTM!

;)

PR #159, opened 2017-01-13
Update vendor script to not use IFS

katcipis:
@tiago4orion The build is breaking when I run vendor...but it also breaks with the same error if I run the old original script, not sure of what went wrong. The error:

```
github.com/NeowayLabs/nash (download)
github.com/chzyer/readline (download)
github.com/NeowayLabs/nash/token
github.com/NeowayLabs/nash/sh
github.com/chzyer/readline
github.com/NeowayLabs/nash/scanner
github.com/NeowayLabs/nash/ast
github.com/NeowayLabs/nash/errors
github.com/NeowayLabs/nash/parser
github.com/NeowayLabs/nash/internal/sh
github.com/NeowayLabs/nash
_/home/katcipis/workspace/go/src/github.com/NeowayLabs/nash/cmd/nash
# _/home/katcipis/workspace/go/src/github.com/NeowayLabs/nash/cmd/nash
./completer.go:34: c.term.PauseRead undefined (type *readline.Terminal has no field or method PauseRead)
exit status 2
```

The vendor directory does not contain the readline project anymore...even with the old script.

### DIFF
--- a/cmd/nash/vendor.sh
+++ b/cmd/nash/vendor.sh
@@ -1,23 +1,30 @@
 #!/usr/bin/env nash
+
 fn vendor() {
-        IFS = ()
-	rm -rf vendor
-	mkdir -p vendor/bin vendor/src vendor/pkg
-	GOPATH <= pwd | xargs echo -n
-	GOPATH = $GOPATH+"/vendor"
-	setenv GOPATH
-	GOBIN = $GOPATH+"/bin"
-	setenv GOBIN
-	go get -v .
-	IFS = ("\n")
-	paths <= ls vendor/src
-	for path in $paths {
-		mv "vendor/src/"+$path vendor/
-	}
-	rm -rf vendor/src vendor/bin vendor/pkg
-	# because nash library is a dependency of cmd/nash
-	# we need to remove it at end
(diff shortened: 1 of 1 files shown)


tiago4orion:
> ./completer.go:34: c.term.PauseRead undefined (type *readline.Terminal has no field or method PauseRead)

Nash vendored readline is modified. We're maintaining our vendor ourselves, not getting updates from upstream anymore because I patched the lib for our use case... That means that we cannot delete the vendor folder anymore... This script became useless now because I'm applying updates in readline manually.

But the script used to work on other projects and your patch to remove IFS variable seems fine.

I'm not sure what's the best way to maintain 3rdparty patched libraries. My hope is delete this readline dependency also and implement the terminal crappy ourselves in the future...

tiago4orion:
I added features and removed the code of everything we doesn't use from readline. Then, we're pretty far from readline master branch to upgrade easy.

tiago4orion:
https://github.com/NeowayLabs/nash/pull/122/files

katcipis:
@tiago4orion I see, thanks for the enlightenment :-)

PR #162, opened 2017-01-18
Add more examples and start reference docs

katcipis:
Having some troubles checking the len of the ARGS :-)


examples/len.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env nash
+
+echo $ARGS
+argslen <= len($ARGS)
+
+test $argslen + "= 1"
```
katcipis: @tiago4orion I cant get this script to work, it must be something stupid and I'm tired :-(. But it is extremely frustrating that a simple comparison on the if is not possible.

doc/reference.md
```diff
@@ -0,0 +1,30 @@
...
+## Script args
+
+To handle script arguments you can just use the ARGS variable,
+that is a list populated with the arguments passed to your script
+when it is executed, like:
+
+```
+echo ""
+echo "acessing individual parameter"
+somearg = $ARGS[0]
```
tiago4orion: In scripts, $ARGS[0] must be the script name. If it's not the case, there's a bug in the cmd/nash.

doc/reference.md
```diff
@@ -0,0 +1,30 @@
...
+echo $somearg
+echo ""
```
tiago4orion: adding newline? Simply `echo` should suffice.

examples/len.sh
tiago4orion: Do you mean the code below?
```sh
if len($ARGS) == "1" {
        echo "one parameter"
}
```
It works...

doc/reference.md
katcipis: It is the script name,no problem there :-)

doc/reference.md
katcipis: did not no, I'm stupid for shell :-(

doc/reference.md
```diff
@@ -0,0 +1,130 @@
...
+#### string
+
+String concatenation is pretty straightforward.
+For example:
+
+```nash
+a = "1"
+b = "2"
+
+echo $a + $b
```
tiago4orion: I know we don't have a style guide yet, but I strongly prefer `$a+$b` instead of `$a + $b`... The latter could trick the reader into thinking the command receives two variables instead of one joined string. I know this could be bullshit, but I prefer the compact form even on Go (gofmt already format this way..).

What do you think?

doc/reference.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Built-in functions
+
+### len
+
+The function **len** returns the length of a list.
+An example to check for the length of a list:
+
+```
+echo "define one list with two elemnts"
```
tiago4orion: typo

doc/reference.md
```diff
@@ -0,0 +1,130 @@
...
+example_list = ()
+echo "appending string 1"
+example_list <= append($example_list, "1")
+echo $example_list
+echo "appending string 2"
+example_list <= append($example_list, "2")
+echo $example_list
+```
+
+## Packages
```
tiago4orion: What do you mean by that? kkk

doc/reference.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Flow control
+
+## Functions
+
+Defining functions is very easy, for example:
+
+```nash
+fn concat(a, b) {
+        return $a, $b
```
katcipis: sure, I'm stupid :P

doc/reference.md
katcipis: Neutral on this, going to do your way ;-)

katcipis:
@tiago4orion Documenting everything is a little daunting for me right now. I think we could merge this and work on it incrementally :-)

tiago4orion:
Top!

PR #168, opened 2017-01-27
Advanced for ranges

tiago4orion:
Enables:

```sh
for i in getfiles("/") {

}
```

and 

```sh
for i in (1 2 3 4 5) {

}
```

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -1952,16 +1952,28 @@ func (shell *Shell) executeFor(n *ast.ForNode) (sh.Obj, error) {
 		shell.looping = false
 	}()
 
-	if n.InVar() == "" {
+	if n.InExpr() == nil {
 		return shell.executeInfLoop(n.Tree())
 	}
 
 	id := n.Identifier()
-	inVar := n.InVar()
+	inExpr := n.InExpr()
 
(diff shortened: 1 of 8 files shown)


vitorarins:
Awesome! :smile:

internal/sh/shell.go
```diff
@@ -1952,24 +1952,36 @@ func (shell *Shell) executeFor(n *ast.ForNode) (sh.Obj, error) {
 		shell.looping = false
 	}()
 
-	if n.InVar() == "" {
+	if n.InExpr() == nil {
 		return shell.executeInfLoop(n.Tree())
```
katcipis: for {} rocks :-). One of the first constructs that I just tried out in nash intuitively and it worked :-)

internal/sh/shell_test.go
```diff
@@ -320,6 +320,7 @@ func TestExecuteCmdAssignment(t *testing.T) {
 	}
 }
 
+// IFS *DO NOT* exists anymore. This tests only assure things works as expected (IFS has no power)
```
katcipis: Cant the test name be changed to reflect better the intention, got a little confused on why this has to stay if IFS has no more power.

tiago4orion: yes, makes sense!

Issue #161, opened 2017-01-18
builtin: implement exit


katcipis:
Implement the Runner interface, like: 

https://github.com/NeowayLabs/nash/blob/master/internal/sh/split.go

To finish the script :-)

vitorarins:
This was implemented by #185

PR #172, opened 2017-02-12
Add support for multi-var assignments

tiago4orion:
This PR will introduce ~two~ three great ideas:

- Command execution getting status in a second variable (deprecating the use of '-' in cmds)
- Functions that return multiple values
- tuple assignment

Then, the following snippets will be valid:

Getting status in a second var:
```sh
ls, err <= ls /dir-do-not-exists

if $err != "0" {
    echo "Directory does not exists"
}
```

Functions with multiple returns:

```sh
fn getusercfg() {
    user, status <= echo -n $USER
    
    if $status != "0" {
            return "", "", "environment variable $USER not set"
    }

    homedir = "/home/"+$user
    homeExists, status <= test -d $homedir

    if $status != "0" {
        return "", "", "home directory not found"
    }
    
    return $user, $homedir, ""
}

fn main() {
     user, home, err <= getusercfg()

     if err != "" {
        echo "current user cfg not found"
        exit(1)
    }

    echo "User:" $user
    echo "Home:" $home
}
```

Tuple assignments:

```sh
a, b = "hello", "world"
echo $a $b
```


ast/node.go
```diff
@@ -99,6 +99,17 @@ type (
...
+	ExecMultipleAssignNode struct {
+		NodeType
+		token.FileInfo
+
+		names   []*NameNode
+		cmd     Node
+		eqSpace int
+	}
+
 	// A CommandNode is a node for commands
```
katcipis: I think in Go comments always start with the name of what is being documented. So here it would be:

// CommandNode is a node for commands

Which by the way is so simple that I'm not even sure if the comment is pulling its own weight :-)

ast/node.go
```diff
@@ -687,6 +701,84 @@ func (n *ExecAssignNode) IsEqual(other Node) bool {
...
 }
 
+// NewExecMultipleAssignNode creates a new node for executing something and storing the
+// result on multiple variables. The assignment could be made using an operating system
+// command, a pipe of commands or a function invocation.
+// It returns a *ExecMultipleAssignNode ready to be executed or error when
+// n is not a valid node for execution.
+func NewExecMultipleAssignNode(info token.FileInfo, names []*NameNode, n Node) (*ExecMultipleAssignNode, error) {
+	if !n.Type().IsExecutable() {
+		return nil, errors.New("NewExecAssignNode expects a CommandNode, PipeNode or FninvNode")
```
katcipis: If this is the error that will be exposed to the user he/she will not understand anything, but I supposed the error will be wrapped better on the call chain, going on reviewing :-)

ast/node.go
```diff
@@ -687,6 +701,84 @@ func (n *ExecAssignNode) IsEqual(other Node) bool {
...
+
+		names:   names,
+		cmd:     n,
+		eqSpace: -1,
+	}, nil
+}
+
+// Name returns the identifier (l-value)
+func (n *ExecMultipleAssignNode) Names() []*NameNode {
+	return n.names
```
katcipis: Why not just expose the Names field ?

ast/node.go
```diff
@@ -687,6 +701,84 @@ func (n *ExecAssignNode) IsEqual(other Node) bool {
...
+}
+
+func (n *ExecMultipleAssignNode) getEqSpace() int      { return n.eqSpace }
```
katcipis: Why not just expose the field ? This field is even private...really not getting why it even exist

ast/node.go
```diff
@@ -687,6 +701,84 @@ func (n *ExecAssignNode) IsEqual(other Node) bool {
...
+func (n *ExecMultipleAssignNode) setEqSpace(value int) { n.eqSpace = value }
```
katcipis: Ok, I will stop talking about that now, you get the idea :-). I imagine that perhaps this is done to maintain consistency with the rest of the code, but to be honest I don't think this is a good path to go on, using this settters/getters stuff. IMHO, sometimes refactoring requires some period where things are inconsistent, but at least we can start changing the direction of the code towards just accessing fields directly.

ast/node.go
tiago4orion: This rule is for documentation of function/methods/consts/etc. When documenting types you can use "a/an" to improve wording (but it's optional).

Something like: 

```go
// A Decoder is used to blah blah blah
type Decoder struct{}
```

But regarding what I wrote for documentation... I agree.. useless hahaha

ast/node.go
tiago4orion: This error will show for people generating code (or ASTs) using `github.com/NeowayLabs/nash/ast` package directly. Our parser builds an `ast.Tree` and invokes this methods to build nodes. Our `parse.Parser` shall not trigger this error because otherwise it's has a bug...

This error will trigger in case you're incorrectly building a nash AST directly, for example, if someone write the honored project of translating bash scripts into nash code :-)

ast/node.go
tiago4orion: I admit this is weird, but it has a reason. Go interfaces are build upon methods, I cannot use a field to satisfy an interface... Maybe I'm handling this in the wrong way. Maybe you can help me get this right.

Take a look in this interface:
https://github.com/NeowayLabs/nash/blob/master/ast/node.go#L36

and here:
https://github.com/NeowayLabs/nash/blob/e6eb8565e7dbece60a71e655f09c7aeafe1de680/ast/node_fmt.go#L80

The idea for that is to align the indentation of several assignments cases when consecutive:

```sh
pathdir            = "/etc/rules.d"
rulegroup          = "admin"
files, status, err <= getrules("/")
```

Even if I add a dumb method called `IAmAssignable` to the interface... I'll still need the private methods `getEqSpace/setEqSpace` because there's no way to use fields when using a interface instance... What do you think?

But thanks for pointing it out because I just saw this new type `ExecMultipleAssignNode` doesn't implement the `assignable` interface... And this was not catch because the missing formatting tests for it.

ast/node.go
tiago4orion: I dislike getters/setters also :-) and much more when the field is private o.o but see my other comment regarding this case.

ast/node.go
tiago4orion: I agree, but see next comment.

ast/node.go
katcipis: Did not know, thanks :-)

ast/node.go
katcipis: I see, good luck to the pour soul who tries to do that :P

ast/node_fmt.go
```diff
@@ -236,6 +236,48 @@ func (n *ExecAssignNode) String() string {
...
+		multi  bool
+		lhs    string
+	)
+
+	for i := 0; i < len(n.names); i++ {
+		name := n.names[i]
+
+		lhs += name.String()
+
+		if i < len(n.names)-1 {
```
katcipis: You could ditch all the logic of building the string incrementally and just create an array of name.String() and in the end call strings.Joing(array, ","). Also I think it is better for the CPU time since strings are immutable, so concatenating on immutable strings is usually not very fast. But I'm not sure if Go does not implement some kind of hack to perform better on concatenations.

ast/node_fmt.go
```diff
@@ -610,11 +652,23 @@ func (n *DumpNode) String() string {
...
+	returnExprs := n.arg
+
+	for i := 0; i < len(returnExprs); i++ {
+		if i == 0 {
+			ret += " "
+		}
+
+		ret += returnExprs[i].String()
+
+		if i < len(returnExprs)-1 {
```
katcipis: another candidate for using strings.Joing

ast/nodetype_string.go
```diff
@@ -4,9 +4,9 @@ package ast
 
 import "fmt"
 
-const _NodeType_name = "NodeSetenvNodeBlockNodeAssignmentNodeExecAssignNodeImportexecBeginNodeCommandNodePipeNodeRedirectNodeFnInvexecEndexpressionBeginNodeStringExprNodeIntExprNodeVarExprNodeListExprNodeIndexExprNodeConcatExprexpressionEndNodeStringNodeRforkNodeRforkFlagsNodeIfNodeCommentNodeFnDeclNodeReturnNodeBindFnNodeDumpNodeFor"
+const _NodeType_name = "NodeSetenvNodeBlockNodeNameNodeAssignmentNodeExecAssignNodeExecMultipleAssignNodeImportexecBeginNodeCommandNodePipeNodeRedirectNodeFnInvexecEndexpressionBeginNodeStringExprNodeIntExprNodeVarExprNodeListExprNodeIndexExprNodeConcatExprexpressionEndNodeStringNodeRforkNodeRforkFlagsNodeIfNodeCommentNodeFnDeclNodeReturnNodeBindFnNodeDumpNodeFor"
```
katcipis: what the hell is that ? :-P

internal/sh/shell.go
```diff
@@ -1529,21 +1540,27 @@ func (shell *Shell) evalExpr(expr ast.Expr) (sh.Obj, error) {
...
 				return nil, errors.NewEvalError(shell.filename,
 					expr,
 					"Function used in"+
 						" expression but do not return any value: %s",
 					fnInv)
+			} else if len(objs) != 1 {
+				return nil, errors.NewEvalError(shell.filename,
+					expr,
+					"Function used in"+
+						" expression but it returns %d values: %d",
```
katcipis: The formatting has two %d but only the len(objs) is passed around, is this right ? This seems to be forcing to functions not be able to be used on expressions when there is more than one parameter involved, so I cant do something like:

```
fn a (x, y) {
}

fn b () {
    return 1, 2
}

a(b())

```

Right ? Is this by design, like this is a bad thing, or because implementing this would be hard ?

internal/sh/shell.go
```diff
@@ -1612,58 +1629,168 @@ func (shell *Shell) concatElements(expr *ast.ConcatExpr) (string, error) {
...
+		assign := v.(*ast.ExecAssignNode)
+		cmd = assign.Command()
+		output, status, err := shell.execCmdOutput(cmd)
 
 		if err != nil {
 			return err
 		}
 
-		if fnValues == nil {
+		// compatibility mode
```
katcipis: makes sense :+1:

internal/sh/shell.go
```diff
@@ -1612,58 +1629,168 @@ func (shell *Shell) concatElements(expr *ast.ConcatExpr) (string, error) {
...
+
+func (shell *Shell) executeExecAssignFn(v ast.Node) error {
+	var (
+		err      error
+		fnValues []sh.Obj
+		cmd      ast.Node
+		names    []*ast.NameNode
+	)
+
+	if v.Type() == ast.NodeExecAssign {
```
katcipis: If both NodeExecAssign and NodeExecMultipleAssign had a "Names" method, where NodeExecAssign is just a special case that always return only one name (slice with len 1) it seems like kind of if could be avoided. But I'm not sure if this does not impact something else I may have missed.

internal/sh/shell.go
```diff
@@ -1612,58 +1629,168 @@ func (shell *Shell) concatElements(expr *ast.ConcatExpr) (string, error) {
...
+		assign.Type() == ast.NodePipe {
+		return shell.executeExecAssignCmd(v)
+	}
+
+	return errors.NewEvalError(shell.filename,
+		assign, "Invalid node type (%v). Expected function call, command or pipe",
+		assign)
+}
+
+func (shell *Shell) executeExecMultipleAssign(v *ast.ExecMultipleAssignNode) error {
```
katcipis: This seems to be identical to executeExecAssign, or I missed something ? (with the exception of the parameter, but this seems like something we can fix with some abstraction)

internal/sh/shell.go
```diff
@@ -1938,7 +2061,15 @@ func (shell *Shell) executeFor(n *ast.ForNode) (sh.Obj, error) {
 	} else if inExpr.Type() == ast.NodeListExpr {
 		obj, err = shell.evalList(inExpr.(*ast.ListExpr))
 	} else if inExpr.Type() == ast.NodeFnInv {
-		obj, err = shell.executeFnInv(inExpr.(*ast.FnInvNode))
+		var objs []sh.Obj
+		objs, err = shell.executeFnInv(inExpr.(*ast.FnInvNode))
+
+		if len(objs) != 1 {
+			return nil, errors.NewEvalError(shell.filename,
+				inExpr, "Functions with multiple returns do not work as for 'in expression' yet: %v", inExpr)
```
katcipis: Fair enough :-)

internal/sh/shell_test.go
```diff
@@ -320,6 +320,72 @@ func TestExecuteCmdAssignment(t *testing.T) {
...
+			`name, err <= echo -n i4k
+                         if $err == "0" {
+                             echo -n $name
+                         }`,
+			"i4k", "",
+			"",
+		},
+		{
+			"list cmd assignment",
+			`name, err2 <= echo "honda civic"
```
katcipis: You really like this car :-), I think it would be easier to understand these if the execTest structs where initialized by name (also easier to extend/refactor on the future). It is kind hard to understand what is each of the strings on execTest.

internal/sh/shell_test.go
```diff
@@ -320,6 +320,72 @@ func TestExecuteCmdAssignment(t *testing.T) {
...
+                         if $err == "0" {
+                             echo -n $l
+                         }`,
+			`0 1 666 3`,
+			"",
+			"",
+		},
+	} {
+		testExec(t,
+			test.desc,
```
katcipis: Why not accept execTest as parameter ? Also, execTest seems like it is a test actually, instead of a description of a test case. Perhaps it could be  execTestCase or execTestDesc ?

internal/sh/split.go
```diff
@@ -157,7 +157,15 @@ func splitByFn(content string, splitFunc sh.Fn) []string {
 			return false
 		}
 
-		result := splitFunc.Results()
+		results := splitFunc.Results()
+
+		// fuck... now this is broken...
+		// remove support of split by func?
```
katcipis: Not sure why it has to be broken, but it is better to break it then give half support to it, I think.

katcipis: I really did not get it, need some help :P

parser/parse.go
```diff
@@ -663,9 +663,9 @@ func (p *Parser) parseAssignment(ident scanner.Token) (ast.Node, error) {
 	it := p.next()
 
 	if it.Type() != token.Assign && it.Type() != token.AssignCmd &&
```
katcipis: It would be good to extract this if with lots of comparisons to a function that expresses better what you are trying to validate. And to avoid if'ing around at all a map can help, if you don't think it is a overkill.

parser/parse.go
```diff
@@ -683,11 +683,67 @@ func (p *Parser) parseAssignment(ident scanner.Token) (ast.Node, error) {
...
+
+		names = append(names, ast.NewNameNode(it.FileInfo, name.Value(), index))
+
+		if it.Type() != token.Comma {
+			break
+		}
+	}
+
+	if it.Type() != token.AssignCmd {
+		// will we support multiple simple assignment (no cmd exec)?
```
katcipis: Something like:

```
a, b, c = "1", "2", "3"
```

Seems like a consistent choice, but I would not die without it.

parser/parse.go
```diff
@@ -813,15 +869,21 @@ func (p *Parser) parseAssignCmdOut(identifier scanner.Token, index ast.Expr) (as
 		return nil, err
 	}
 
-	var name *ast.NameNode
+	if len(identifiers) == 0 {
+		// should not happen... pray
+		panic("internal error parsing assignment")
```
katcipis: Isn't there any other info that could make this panic a little more useful than "shit just get serious" ? :-)

parser/parse_test.go
```diff
@@ -214,6 +214,54 @@ func TestBasicAssignment(t *testing.T) {
...
+
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	ln.Push(assign)
+	expected.Root = ln
+
+	parserTestTable("multiple cmd assignment", `test, status <= ls`, expected, t, true)
```
katcipis: I find cool to keep the testing.T parameter always as the first one on test functions, so it looks consistent with go's testing package

parser/parse_test.go
```diff
@@ -214,6 +214,54 @@ func TestBasicAssignment(t *testing.T) {
...
+
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	ln.Push(assign)
+	expected.Root = ln
+
+	parserTestTable("multiple cmd assignment", `test[0], status <= ls`, expected, t, true)
```
katcipis: Testing two different scenarios on the same test is usually not a very good idea. Or you are actually testing two different things and they deserve a test for their own, or one of the tests is not necessary. On this case it seems you are testing a really important case, that is different from the previous one.

parser/parse_test.go
```diff
@@ -214,6 +214,54 @@ func TestBasicAssignment(t *testing.T) {
...
+}
+
 func TestParseInvalidIndexing(t *testing.T) {
 	// test indexed assignment
 	parser := NewParser("invalid", `test[a] = "a"`)
```
katcipis: Smells like multiple tests embedded on the same. Perhaps some table testing using subtests ? :-)

ast/node.go
tiago4orion: > Formatting is cool, but AFAIK is not syntactic/semantic relevant.

Every node must have a string representation, don't you think? Then all this formatting stuff came up.. The output of the parser is an `*ast.Tree` and being possible to print the AST is really useful for debugging purposes. But maybe what you're saying is to just move this code to somewhere else? Maybe create a `ast/printer` package and put this logic inside?

ast/node.go
```diff
@@ -1289,12 +1381,12 @@ func NewReturnNode(info token.FileInfo) *ReturnNode {
 }
 
 // SetReturn set the arguments to return
-func (n *ReturnNode) SetReturn(a Expr) {
+func (n *ReturnNode) SetReturn(a []Expr) {
```
tiago4orion: got me =D

I'm not entirely sure when I started doing this way, but I think now it was a bad idea ... I remember that I'd to put some properties wrapped in setters because some of them required a kind of validation (or other thing) before the assignment... and then this became the rule for consistency in the API... 

But maybe I should allow the user to set what they want in the struct, but validate only when used?
I dislike this also, and I'll fix it in this PR =) Thanks.

ast/node.go
```diff
@@ -1315,15 +1407,20 @@ func (n *ReturnNode) IsEqual(other Node) bool {
...
+		return false
 	}
 
-	if n.arg != nil {
-		return n.arg.IsEqual(o.arg)
+	for i := 0; i < len(n.arg); i++ {
+		arg := n.arg[i]
+		oarg := o.arg[i]
+
+		if arg != nil && !arg.IsEqual(oarg) {
```
tiago4orion: Nice! Thanks!

ast/node_fmt.go
tiago4orion: Yeah! I was doing this way for other nodes... https://github.com/NeowayLabs/nash/blob/master/ast/node_fmt.go#L347

I worked hard in the last weekend to get the taste of this features that didn't stop to think or re-read the current implementation... sorry =(

I'll fix this!

ast/node_fmt.go
tiago4orion: thanks

ast/nodetype_string.go
tiago4orion: https://blog.golang.org/generate
https://godoc.org/golang.org/x/tools/cmd/stringer

Rob Pike tool to generate the `String()` method for consts.

internal/sh/shell.go
tiago4orion: Thanks, you're right, the last %d should be %v and I'm missing a parameter.

Regarding the semantics of this eval: Yeah, your snippet is not valid. Because:

1- Expression can exists in other places where this doesn't makes sense and then make this work only in funcalls is weird... The current implementation requires that expressions yield a single object. Eg.: In case for the `b` function you declared, this snippet doesn't makes sense: `if b() == ?? {}`

2- For me (personal opinion) the statement `a(b())` misguide the user... It appears that function `a` receives one parameter ... but no.. And things start to be weird with things like: `a(b(), c, d())` where `b` and `d` has multiple returns.. If nash were a static compiled language the mistakes of changing the number of returns in `b` and not breaking the uses of `a(b())` would be easy to catch... but nash is dynamic interpreted..
3- Go is static compiled and this doens't works also.

internal/sh/shell.go
tiago4orion: I'd asked myself the same question when wrote this line of code hahaha I didn't know the answer at that time also  and decided to *not think* and go straight forward... Now I think the answer is *yes*. Thanks

internal/sh/shell.go
tiago4orion: Agreed!

internal/sh/shell_test.go
tiago4orion: LoL

Regarding instantiating execTest using the field names, yeah, better idea =)

internal/sh/shell_test.go
tiago4orion: > Why not accept execTest as parameter ?

I'm an idiot :-)

internal/sh/split.go
tiago4orion: It's already half-supported in this implementation. I was complaining only because I dislike the idea of contract by documentation... the builtin `split` function support split strings by user-defined functions (but only for single return...).

parser/parse.go
tiago4orion: same here.. I would like for consistency.. but not I would use so much..

parser/parse.go
tiago4orion: I could the remove the `if` entirely. An out-of-bounds panic from Go runtime is more useful than this :-) What do you think?

parser/parse_test.go
tiago4orion: makes sense

parser/parse_test.go
tiago4orion: sorry, I'm lazy kkk

parser/parse_test.go
tiago4orion: Yeah, refactor this (and others) tests are on my todo list =)

tiago4orion:
@katcipis Thank you! Really appreciate your help!

I'll try to improve with your suggestions in the next weekend.

ast/node.go
katcipis: Maybe create a ast/printer package and put this logic inside? = Exactly :+1: 

I'm not arguing to having a printer/format capability, just arguing with mixing up formatting and AST code all together. It seems better to have this:

```
printer (depends on) -> ast (only syntatic stuff)
```

Instead of:

```
ast(lots of formatting stuff with the syntatic stuff and with parsing and with a lot of other already complex stuff).
```

The first hint for me is that although formatting is useful it is not related to the AST. I agree with the node being able to represent itself as a string, but I thinking formatting as a whole is too much (at least taking a look at the code it seemed like this).

ast/node.go
katcipis: If there is validation involved, we can think about it. This one is a no brainer since there is no validation going on.

If there is validation it would be cool to fail as fast as possible, if you are going to fail too far ahead on the road, it may get confusing, it would be a good case of a "Setter".

ast/node_fmt.go
katcipis: Stuff is already looking great, chill ;-)

internal/sh/shell_test.go
katcipis: uaheuahueahuehauehauheuhaeuahaueh

when you writing tons of stuff to get shit working this kind of stuff happens, idiots can't add cool features on awesome languages ;-)

internal/sh/split.go
katcipis: well, in this case it seems like the contract of the function is a single return, that must be a list...can't see why this would be a problem o.O

parser/parse.go
katcipis: @tiago4orion have you seen this comment ?

parser/parse.go
katcipis: later then :-)

parser/parse.go
katcipis: If it is really exceptional case, perhaps it would be better. If you can augment it with domain specific information it is usually better, but if there is nothing to be added, allow the runtime to do its job :+1:

parser/parse_test.go
katcipis: You are lazy for testing, and thirsty for hacking :-). Perhaps someday you will see testing as hacking too :-P

parser/parse_test.go
katcipis: I can help with that ;-)

ast/node.go
tiago4orion: :+1:

ast/node.go
tiago4orion: but an API that has direct field access in some cases and requires setters in others... is not nice.

parser/parse.go
tiago4orion: I haven't seen it!

I cannot think in a good function name right now... I'll revisit this. Thanks

ast/node.go
tiago4orion: This doesn't work when the object is an interface :-(

tiago4orion:
```go
type Stringer interface {
    String() string 
}

var s Stringer

s.String() // panic
```

ast/node.go
katcipis: I would argue that the setter is not only a setter actually, it does more stuff than setting a field, I even try to come up with better names for it than just "set". I'm not sure if this violates the idea of being uniform, lack of uniformity would be to mix setter and field accessing on the same cases. When some pre validation is required to enforce invariants on a object it does not seem to me as just a setter, although sometimes it is hard to come up with a better name :-).

ast/node.go
katcipis: Actually, if you have a concrete type that is nil and pass it as a implementation for the interface it works:

https://play.golang.org/p/Rt3Sof1Xcd

But I see your point, on the case you showed here there is no type descriptor initialized on the Stringer interface, it has no way to decide which method to call. On the snipper I provided I initialized it with a valid type descriptor, but with a nil value pointer, in that case it can dispatch the method passing nil as a parameter.

It seems to me that this is not related to having a method that can accept a nil value, this is related to having a interface that has no type descriptor attached on it.

parser/parse.go
katcipis: :+1:

ast/node.go
tiago4orion: I'll create an issue to refactor this into `ast/printer` in the future because the change turns out to be very complex and this will mess this PR that's related to multiple returns. Ok ?

parser/parse.go
```diff
@@ -1223,78 +1275,86 @@ func (p *Parser) parseReturn(retIt scanner.Token) (ast.Node, error) {
...
+			p.ignore()
 
-			if err != nil {
-				return nil, err
-			}
+			var values []ast.Expr
 
-			values = append(values, arg)
-		}
+			for valueIt = p.peek(); valueIt.Type() != token.RParen && valueIt.Type() != token.EOF; valueIt = p.peek() {
```
vitorarins: No problem here but I think it's a little hard to read. Maybe create a function would help?

tiago4orion: I'll try to improve this!

Thanks!

tiago4orion:
@katcipis @vitorarins Can you guys take a look?

Everything is working here, but maybe there's cases that I'm forgetting to test..

In the PR there's a partial implementation of feature below also:

```sh
a, b = "hello", "world"
c, d, r = "1", "2", "3"
```
Added this for consistency, but it was implemented in the lexer and evaluator, but not in the parser yet ;)

tiago4orion:
Tests failed because /use/bin/cat doesn't exists in Travis image... Will fix that ASAP.

tiago4orion:
Added more missing tests.

@katcipis @vitorarins Can we merge this?

tiago4orion:
Oops... It depends on #161 ...

tiago4orion:
@katcipis Have you forgotten this?

katcipis:
@tiago4orion yep, sorry :-). Looking at it ASAP

parser/parse.go
```diff
@@ -177,18 +177,17 @@ func (p *Parser) parseStatement() (ast.Node, error) {
...
 
 	// statement starting with ident:
-	// - fn invocation
+	// - fn call
 	// - variable assignment
 	// - variable exec assignment
 	// - Command
 
-	if (it.Type() == token.Ident || it.Type() == token.Variable) && next.Type() == token.LParen {
+	if isFuncall(it.Type(), next.Type()) {
```
katcipis: much better :+1:

ast/doc_test.go
```diff
@@ -0,0 +1,50 @@
+package ast_test
+
+import (
+	"fmt"
+
+	"github.com/NeowayLabs/nash/ast"
+	"github.com/NeowayLabs/nash/token"
+)
+
+func Example_AssignmentNode() {
```
katcipis: Examples, cool :-)

fuzz.go
```diff
@@ -2,14 +2,16 @@
```
katcipis: Fuzzing action lol

internal/sh/builtin.go
```diff
@@ -89,7 +89,6 @@ func (f *builtinFn) SetStdin(r io.Reader)  { f.stdin = r }
 func (f *builtinFn) SetStderr(w io.Writer) { f.stderr = w }
 func (f *builtinFn) SetStdout(w io.Writer) { f.stdout = w }
 func (f *builtinFn) StdoutPipe() (io.ReadCloser, error) {
```
katcipis: Trying to abstract functions just as commands does not seem to have helped bash, a lot of things are very ugly/odd in shell because of that. So I'm not sure if we have to worry that some things are different in functions

parser/parse.go
```diff
@@ -683,11 +688,47 @@ func (p *Parser) parseAssignment(ident scanner.Token) (ast.Node, error) {
...
 	}
 
-	if it.Type() == token.Assign {
-		return p.parseAssignValue(ident, index)
+	names := []*ast.NameNode{
+		ast.NewNameNode(ident.FileInfo, ident.Value(), index),
+	}
+
+	if it.Type() != token.Comma {
+		goto assignOp
```
katcipis: Why this can't be a function call and a return ? The functions is already a little big, goto's are not making any favors to the function :-)

parser/parse.go
```diff
@@ -739,47 +780,55 @@ func (p *Parser) parseList(tok *scanner.Token) (ast.Node, error) {
...
+		var (
+			value ast.Expr
+			err   error
+		)
 
-	if it.Type() == token.Variable || it.Type() == token.String {
-		value, err = p.getArgument(false, true)
+		if it.Type() == token.Variable || it.Type() == token.String {
+			value, err = p.getArgument(false, true)
+		} else if it.Type() == token.LParen { // list
```
katcipis: Usually if I need a if else I go for a switch, specially because Go switch is so more awesome than C switches :-)

parser/parse.go
```diff
@@ -739,47 +780,55 @@ func (p *Parser) parseList(tok *scanner.Token) (ast.Node, error) {
...
-	}
 
-	if p.peek().Type() == token.Semicolon {
 		p.ignore()
 	}
 
-	var name *ast.NameNode
+	if len(values) == 0 {
+		return nil, newParserError(p.peek(), p.name, "Unexpected token %v. Expecting VARIABLE, STRING or (", p.peek())
+	} else if len(values) != len(names) {
```
katcipis: No need for else or else if when the if statement returns :-)

parser/parse.go
```diff
@@ -813,15 +862,12 @@ func (p *Parser) parseAssignCmdOut(identifier scanner.Token, index ast.Expr) (as
...
 
-	var name *ast.NameNode
-
-	if index != nil {
-		name = ast.NewNameNode(identifier.FileInfo, identifier.Value(), index)
-	} else {
-		name = ast.NewNameNode(identifier.FileInfo, identifier.Value(), nil)
+	if len(identifiers) == 0 {
+		// should not happen... pray
+		panic("internal error parsing assignment")
```
katcipis: Is there any more context that can be given to help someone debugging this ? Like some hint on what could possibly be wrong to cause identifiers to be 0 ?

parser/parse.go
```diff
@@ -1201,100 +1247,115 @@ func (p *Parser) parseDump(dumpIt scanner.Token) (ast.Node, error) {
...
 	// return <fn name>()
-	if valueIt.Type() != token.Semicolon &&
-		valueIt.Type() != token.RBrace &&
-		valueIt.Type() != token.Variable &&
-		valueIt.Type() != token.String &&
-		valueIt.Type() != token.LParen &&
-		valueIt.Type() != token.Ident {
-		return nil, newParserError(valueIt, p.name,
+	// return "val1", "val2", $val3, test()
+	if tok.Type() != token.Semicolon &&
```
katcipis: What about a map of the valid possible types and just check if tok.Type() is inside the valid set of values ? This kind of inverse logic (testing what it can't be) don't make it very explicit what it can actually be, the developer has to run set operations on its head to come up with the symmetric difference of the whole sets of values to then come up with what is valid, or the developer could trust your error message :-) (I don't, just trust code :-P).

parser/parse.go
```diff
@@ -1201,100 +1247,115 @@ func (p *Parser) parseDump(dumpIt scanner.Token) (ast.Node, error) {
...
-	if valueIt.Type() == token.Semicolon {
-		p.ignore()
-		return ret, nil
-	}
+	var returnExprs []ast.Expr
 
-	if valueIt.Type() == token.RBrace {
-		return ret, nil
-	}
+	for {
```
katcipis: All good things in life starts with a while true :-)

katcipis: the loop is really big, extracting some of the logic to more functions could improve readability, if you are up to it :-)

parser/parse.go
```diff
@@ -1408,3 +1469,21 @@ func isValidArgument(t scanner.Token) bool {
 
 	return false
 }
+
+func isFuncall(tok, next token.Token) bool {
+	return (tok == token.Ident || tok == token.Variable) &&
+		next == token.LParen
+}
+
+func isAssignment(tok token.Token) bool {
```
katcipis: this kind of predicates are great :-). I miss having booleans on nash to be able to do this.

spec.ebnf
```diff
@@ -5,11 +5,13 @@ program = { statement } .
...
-varSpec      = ( list | string ) .
-string       = stringLit | ( stringConcat { stringConcat } ) .
-assignCmdOut = identifier "<=" ( command | fnInv ) .
+varDecl        = assignValue | assignCmdOut .
+assignValue    = identifierList "=" varSpecList .
+identifierList = identifier [ "," identifierList ] .
+varSpecList    = varSpec [ "," varSpecList ] .
+varSpec        = ( list | string ) .
+string         = stringLit | ( stringConcat { stringConcat } ) .
+assignCmdOut   = identifier "<=" ( command | fnInv ) .
```
katcipis: This syntax does not contemplates:

```
out, status <= cmd
```

Or am I wrong ?

internal/sh/builtin.go
tiago4orion: I'm not changing the language related to this. We already made the choice to make functions work like command and this is the feature behind aliases/bindfn.

katcipis: You already have, passing parameters to a function and to a command is not the same :-). Bash took it to the level where the parameters are read just like a command would.

katcipis: You where the one that explained this to me actually :-P

parser/parse.go
tiago4orion: The goto makes sense to the reader because if it's not a Comma then skip to rhs of the assignment. The function isn't big.. It has 70 lines and gotos are helpful in parser to avoid several meaningless functions.
I'm trying to fit some parser logic in single functions whenever possible to make reading easier, but I do consider splitting in functions when getting bigger.
The problem is the *goto* or the length of the function? If it's the latter I can create a function called `parseAssignNames` and put the for loop just below inside to get the assignment names.

I do think gotos are ok in this cases, but I can change if you think it complicates the readability.

parser/parse.go
tiago4orion: yes, thanks ;-)

parser/parse.go
tiago4orion: yeah, thanks =)

parser/parse.go
katcipis: Your call ;-)

parser/parse.go
tiago4orion: It shouldn't happen because at this parser stage `identifiers` must have one or more elements, otherwise the `parserAssignment` function has a bug and didn't validates the number of identifiers parsed. Do you think there's something more to say in the message?

katcipis: "parserAssignment function has a bug and didn't validates the number of identifiers parsed" ? :-)

You avoided the slicing panic, but the message does not help a lot (just as the slice panic)

parser/parse.go
tiago4orion: I think the idea is good. I'll try to improve that. Thanks =)

parser/parse.go
tiago4orion: :-)

spec.ebnf
tiago4orion: yesss, I forgot! thanks ;)

katcipis: :heart:

internal/sh/builtin.go
katcipis: sorry, misunderstood what you mean't here :heart:

tiago4orion: Ok, I misunderstand your first comment. When I say *already made the choice to  make functions work like command* I mean functions that has stdin, stdout and stderr and can be used in pipelines just as commands.
But yes, bash and nash are completely different syntactic and semantically in various ways.

PR #174, opened 2017-02-16
fix formatting of carrige return

### LINKED ISSUE #173: ast/formatter: carriege returns dropped by nashfmt

tiago4orion:
Closes #173

### DIFF
--- a/ast/node_fmt_test.go
+++ b/ast/node_fmt_test.go
@@ -8,7 +8,7 @@ import (
 
 func testPrinter(t *testing.T, node Node, expected string) {
 	if node.String() != expected {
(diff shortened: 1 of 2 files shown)


ast/node_fmt_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"testing"
+
+	"github.com/NeowayLabs/nash/token"
+)
+
+func testPrinter(t *testing.T, node Node, expected string) {
+	if node.String() != expected {
+		t.Errorf("Values differ: '%s' != '%s'", node.String(), expected)
```
katcipis: You can just pass node, no ? (no need to call .String())

tiago4orion: =)

PR #176, opened 2017-02-18
Refactor builtin functions

katcipis:
Some refactoring preparing the way for the exit implementation. I had some trouble extracting what was exactly the implementation of a builtin function and what was inner stuff to implement the Runner interface with a lot of stdout/err/in stuff. It was a little odd at first because AFAIK functions cant even read stuff from stdin, like:

```
cat file | function()
```

So the fact that builtin functions had a lot of stdout stuff on them was really odd. I can't pinpoint a bigger refactoring right now, but I did what I could to at least make it easier to write builtin functions. Did it by isolating the builtin functions from the internal.sh part of the design, and there is a mapper that gets the isolated builtin function and transforms it on the internal.sh.Fn thing, that for now still seems a little odd to me.

Feel free to think at this as completely garbage :-)

It is a WIP because I have to add split tests, I removed the ones that existed because it seems that it will be much easier to just use nash code to test split, just like the len and append tests :-)


katcipis:
Added some tests for split, still missing tests for splitting by function.

tiago4orion:
Very nice!! I'll take a look right now!

```sh
cat file | function()
```

Do you have any strong feelings about adding support for this?
Currently it's not supported only because I'm not sure it's a good idea..

To functions be used in pipes it will requires that they read stuff explicitly from /dev/stdin and outputs to /dev/stdout... Something like that:

```sh
fn filter(str) {
    cat /dev/stdin | grep $str 
}

string /usr/bin/cat | filter("main")
```

But in other words... not all functions should be used in pipes..
What do you think?

katcipis:
No feelings on piping functions right now, I just mentioned that because it was odd that the builtin functions had stdin/out/err logic on them...exactly because they do not interact with that. I had the feeling of some kind of abstraction broken/leaking, which was the motivation for this refactoring :-)

Elixir has piping on functions, but it does not work as shell commands pipes and we have a trillion of better stuff to implement before something like that.

internal/sh/builtin/len_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if "10" != string(out.Bytes()) {
+		t.Errorf("String differs: '%s' != '%s'", "10", string(out.Bytes()))
+		return
+	}
+
+	// Having to reset is a bad example of why testing N stuff together
+	// is asking for trouble :-).
+	out.Reset()
```
tiago4orion: yeah, better to create a table test in the future.. creating a new `bytes.Buffer` to each test iteration.

internal/sh/builtin/loader.go
```diff
@@ -0,0 +1,22 @@
+package builtin
+
+import "github.com/NeowayLabs/nash/sh"
+
+// Fn is the contract of a built in function, that is simpler
+// than the core nash Fn.
+type Fn interface {
```
tiago4orion: very cool :-)

internal/sh/builtin/len_test.go
katcipis: It is actually a good example of why it is bad, my comment has its logic wrong :P

internal/sh/builtin/split.go
```diff
@@ -0,0 +1,118 @@
...
+
+		err = splitFunc.Wait()
+
+		if err != nil {
+			return false
+		}
+
+		result := splitFunc.Results()
+
+		if result.Type() != sh.StringType {
```
tiago4orion: Just realized I'd not checked for nil value here.. o.O

PoC:

```sh
λ> a = "hello world"
λ> fn b() { echo "oops" }
λ> c <= split($a, $b)
oops
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x53054d]

goroutine 55 [running]:
panic(0x610200, 0xc42000c160)
	/home/i4k/.gvm/gos/go1.7/src/runtime/panic.go:500 +0x1a1
github.com/NeowayLabs/nash/internal/sh.splitByFn.func1(0xc400000068, 0xb)
	/home/i4k/.gvm/pkgsets/go1.7/global/src/github.com/NeowayLabs/nash/internal/sh/split.go:162 +0x17d
strings.FieldsFunc(0xc4203e0330, 0xb, 0xc42003be90, 0xc4201043a0, 0x0, 0x0)
	/home/i4k/.gvm/gos/go1.7/src/strings/strings.go:315 +0x7a
github.com/NeowayLabs/nash/internal/sh.splitByFn(0xc4203e0330, 0xb, 0x934f60, 0xc4201b79f0, 0x0, 0x0, 0x0)
	/home/i4k/.gvm/pkgsets/go1.7/global/src/github.com/NeowayLabs/nash/internal/sh/split.go:173 +0x5f
github.com/NeowayLabs/nash/internal/sh.(*SplitFn).run(0xc420084100, 0x0, 0x0)
	/home/i4k/.gvm/pkgsets/go1.7/global/src/github.com/NeowayLabs/nash/internal/sh/split.go:58 +0x39a
github.com/NeowayLabs/nash/internal/sh.(*SplitFn).Start.func1(0xc420084100)
	/home/i4k/.gvm/pkgsets/go1.7/global/src/github.com/NeowayLabs/nash/internal/sh/split.go:77 +0x2f
created by github.com/NeowayLabs/nash/internal/sh.(*SplitFn).Start
	/home/i4k/.gvm/pkgsets/go1.7/global/src/github.com/NeowayLabs/nash/internal/sh/split.go:79 +0x75
ERROR: exit status 2
```

I'll open an issue for that!

katcipis: Sorry, I was focused on the design and the refactor, did not validate the current code, just wanted to make it fit on the new design =/ (my mind is extremely limited and serialized)

internal/sh/shell.go
```diff
@@ -424,21 +425,17 @@ func (shell *Shell) String() string {
...
-	shell.Setvar("append", sh.NewFnObj(appendfn))
-
-	splitfn := NewSplitFn(shell)
-	shell.builtins["split"] = splitfn
-	shell.Setvar("split", sh.NewFnObj(splitfn))
-
-	chdir := NewChdir(shell)
-	shell.builtins["chdir"] = chdir
-	shell.Setvar("chdir", sh.NewFnObj(chdir))
+	for name, builtinfn := range builtin.Load() {
```
tiago4orion: much much more pleasant =)

internal/sh/builtin/testdata/split.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env nash
```
tiago4orion: Did you forget to commit the Go counterpart for this test?

tiago4orion:
Regarding functions having stdin/stdout/stderr it's required to bindfn'ed functions in pipes (this DO works):
When we create a bind of a function to some command name, this function can really be used as a command (inclusive in pipes):

the below works:
```sh
λ> fn grep(pat) {
    cat /dev/stdin | grep $pat 
}
bindfn grep mygrep
λ> cat /etc/passwd | mygrep i4k
i4k:x:1000:1000:i4k:/home/i4k:/home/i4k/.gvm/pkgsets/go1.7/global/bin/nash
```

But only in interactive mode (because bindfn is disabled in scripts).

That's why you see stdin/stdout stuff in functions. The next step could be adding the full support of it in pipes, in the form:

```sh
cat /etc/passwd | mygrep("i4k")
```

But I'm not sure about that... only because people will need to decide if they must create pipeable functions or "normal" functions (that receive arguments and return values).

internal/sh/builtin/split.go
tiago4orion: No problem. I don't think is cool fix a issue while refactoring anyway :-)

internal/sh/builtin.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"io"
+
+	"github.com/NeowayLabs/nash/errors"
+	"github.com/NeowayLabs/nash/internal/sh/builtin"
+	"github.com/NeowayLabs/nash/sh"
+)
+
+type (
+	//builtinFn maps a built in function to a nash sh.Fn
```
tiago4orion: missing the space between '//' and comment here.

internal/sh/builtin.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (f *builtinFn) String() string {
+	return fmt.Sprintf("<builtin function %q>", f.Name())
+}
+
+func (f *builtinFn) SetArgs(args []sh.Obj) error {
+	return f.fn.SetArgs(args)
+}
+
+func (f *builtinFn) SetEnviron(env []string) {
```
tiago4orion: if this function isn't used... better to drop it =)

tiago4orion: I just checked the current usage.
Why `SetEnviron` is in the interface?

We created an abstraction that involves processes (spawned by commands) and funcalls. The thing is: sub-processes needs to inherit parent environment variables, but funcalls no (because they run in the same nash process). But remember that this can change in the future when we add concurrency.. then funcalls will need to inherit parent environ if executed in background.

internal/sh/builtin.go
```diff
@@ -0,0 +1,97 @@
...
+	// do nothing
+	// terrible design smell having functions that do nothing =/
+}
+
+func (f *builtinFn) SetStdin(r io.Reader)  { f.stdin = r }
+func (f *builtinFn) SetStderr(w io.Writer) { f.stderr = w }
+func (f *builtinFn) SetStdout(w io.Writer) { f.stdout = w }
+func (f *builtinFn) StdoutPipe() (io.ReadCloser, error) {
+	// Not sure this is a great idea, for now no builtin function uses it
```
tiago4orion: this will need *only* if we add support for function in pipes...

internal/sh/builtin/doc.go
```diff
@@ -0,0 +1,2 @@
+//builtin is where all built in functions resides
```
tiago4orion: missing the space between '//' and comment here.

internal/sh/builtin/len_test.go
tiago4orion: I got it!

internal/sh/builtin/testdata/split.sh
katcipis: I did, thanks :-P

internal/sh/builtin/split.go
katcipis: Me neither, but I could have noticed and opened the issue. But designing is hard enough to take all my CPU (single core)

internal/sh/shell.go
katcipis: ;-)

internal/sh/builtin.go
katcipis: fixed

internal/sh/builtin.go
katcipis: There is a lot of code that really seemed necessary because of this common abstraction between functions and commands. To be honest this is kind of new to me so no opnions  formed yet :-). About the pipe thing, I really don't have a problem with functions not having a streamed pipe between them, we have better syntactic support since it is the same language and the same process. Streamed pipes are great to integrate different processes that can be written in different languages, but seems too much effort when you want to integrate functions. Would need more shell scripting in my life to see things differently

katcipis:
@tiago4orion added the split by func test and fixed the CI versions.

PR #185, opened 2017-02-25
Add exit

katcipis:
Here goes nothing. On the tests I explored some linux oddities, pretty fun, but we can remove those tests if it is not useful :-)


internal/sh/builtin/exit_linux_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		"statusIsUnsigned": {
+			script: "./testdata/exit.sh",
+			status: "-1",
+			result: "255",
+		},
+		"statusOverflow": {
+			script: "./testdata/exit.sh",
+			status: "666",
+			result: "154", // Why ? For the glory of satan of course :-)
```
tiago4orion: hahahahaha seriously?

internal/sh/builtin/exit_linux_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+	}
+
+	//WHY: Not sure this is a great idea, but we need to exec with the
```
tiago4orion: I do think the idea is good. Better to remove the discussion from code and put it here :-)

internal/sh/builtin/exit_linux_test.go
```diff
@@ -0,0 +1,79 @@
...
+			err := cmd.Run()
+			if err == nil {
+				if desc.status == "0" {
+					return
+				}
+				t.Fatalf("expected error for status: %s", desc.status)
+
+			}
+			if exiterr, ok := err.(*exec.ExitError); ok {
+				if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
```
tiago4orion: this `if` requires an `else` also.. otherwise tests will incorrectly succeed.

internal/sh/builtin/exit_linux_test.go
katcipis: The only kind of comments I enjoy to left on code is the why, code will never tell you why, just what and how. But I dont explain everything, only what is odd. If in this case you think people won't get confused I can remove it.

internal/sh/builtin/exit_linux_test.go
katcipis: thanks :-)

internal/sh/builtin/exit_linux_test.go
katcipis: yep :-P

ppizarro: kkkkkkkk The funny thing is that the os.Exit asks for an integer argument.
I tested in C and although the return value of the main function is integer type the same happens.

ppizarro:
Exit causes the current program terminates immediately but deferred functions are not run.
Can this cause any problem?

internal/sh/builtin/exit_linux_test.go
tiago4orion: To represent 666 in binary requires at least 10bits and the value is 1010011010. The exit syscall expects 1 byte... Using only the least significant 8 bits of 666 yields 154 hehe
makes sense, but WHY Go doesn't uses uint8 for os.Exit is a mistery.

internal/sh/builtin/exit_linux_test.go
tiago4orion: got it, makes sense now =)

tiago4orion:
@ppizarro In the current nash implementation *I think* the answer is **no**. Every nash statement is self-contained (it doesn't leave things to be cleaned in the future and things like that) then when `exit` statement is evaluated every file opened by last statement *should* be already closed, and so on. It's safe because nash interpreter is serial. 

We'll start working in the concurrency soon and then this will be an issue.

Thanks for pointing out :-)

katcipis:
@tiago4orion I'm a little torn on how to handle the unsupported platform thing. There is two core ideas:

* build breaks
* runtime breaks

I like breaking the build, but the error message today is horrible, and I can't find a way to force a good compilation error, like "there is no exit for this plataform".

So I implemented a crappy way here and pushed it, it will give runtime errors. But I'm not sure this is the way.

katcipis:
@ppizarro I think when you call Go's Exit it will not execute deferreds either.

http://stackoverflow.com/questions/27629380/how-to-exit-a-go-program-honoring-deferred-calls

In Go you can only terminate goroutines guaranteeing that, but to terminate the process you can't guarantee that and I think in nash, to guarantee exit semantics, even with concurrency the behaviour will be the same.

If it is a problem we need to implement some other thing, like runtime.Goexit, but that is for the future.

tiago4orion:
@katcipis I think @ppizarro was referring to interpreter issues of abruptly interrupting the whole program. I must say that `exit` is something that was planned since the beginning and because of that some features are limited. As an example, there's no background copying between file descriptors (`cmd >2&1`) in command redirections to avoid data corruption in such cases. 

To make sure everything is fine I looked for problems in the source code again and found 1 (one) issue candidate. Details below:

When nash is running in interactive mode (cli) it needs to setup the _line discipline_ of terminal into `raw/cooked mode` (instead of default linefeed oriented) using _ioctl/termios_ calls and this must be restored when nash exits. If not restored, in case of nash exiting abruptly, the parent process attached to same terminal (for example `bash`), could behave oddly (characters not echoed, or auto completion not working anymore, and so on). In current implementation the restore is made in a deferred function in `cmd/nash/cli.go`:

[defer rline.Close()](https://github.com/NeowayLabs/nash/blob/master/cmd/nash/cli.go#L74)
that calls [readline.go:250](https://github.com/NeowayLabs/nash/blob/master/cmd/nash/vendor/github.com/chzyer/readline/readline.go#L250) and then [terminal.go:227](https://github.com/NeowayLabs/nash/blob/master/cmd/nash/vendor/github.com/chzyer/readline/terminal.go#L227).

But.. I do think this is a minor problem (or no problem at all) in our case. Scripts don't run interactive and the only issue could be .nash configurations issuing `exit` in case of errors... And if that happens, a simple `reset` in terminal solves the problem...

Another thing to have in mind is that this could already happen today if someone kill nash using SIGINT (for SIGKILL there's no way to restore...). I google about how bash handles that and apparently it restores the terminal state in case of `exit`, but not for SIGINT...

tiago4orion:
@katcipis Regarding `exit` implementation.. it must work on _all_ plataforms that nash supports (linux, osx, windows and plan9).

I asked you to develop first for linux to we discuss how handle this in a portable way. This must be merged only when we support all plataforms :/ What do you think?

internal/sh/builtin/exit_linux.go
```diff
@@ -0,0 +1,55 @@
+package builtin
```
tiago4orion: This implementation is portable :-)

We're using `os.Exit(int)`

internal/sh/builtin/exit_linux_test.go
```diff
@@ -0,0 +1,81 @@
...
+			cmd := exec.Command(projectnash, desc.script, desc.status)
+			err := cmd.Run()
+			if err == nil {
+				if desc.status == "0" {
+					return
+				}
+				t.Fatalf("expected error for status: %s", desc.status)
+
+			}
+			if exiterr, ok := err.(*exec.ExitError); ok {
```
tiago4orion: This section of code is non-portable, but works on every POSIX platform (including windows).
Better to put this section inside a `getstatus` function in a separate files with build flags. This implementation could be inside `exit_posix_test.go`and with the build flags below:

```go
// +build darwin dragonfly freebsd linux nacl netbsd openbsd solaris windows
```

Then, the only missing implementation is plan9.

tiago4orion: I tested here and the plan9 implementation could be something like:

```go
func getstatus(err error) (string, error) {
	if exiterr, ok := err.(*exec.ExitError); ok {
		if status, ok := exiterr.Sys().(syscall.Waitmsg); ok {
			statuscode := status.ExitStatus()
			return strconv.Itoa(statuscode), nil
		}
	}

	return "", err.Error()
}
```

internal/sh/builtin/exit_linux_test.go
vitorarins: I think the comment should stay, but I could not understand the "why we need to exec with the code under test nash" just by reading the comment.

Perhaps you could explain the need of calling `projectnash` by showing the side effects you have when running the exit syscall.

katcipis:
On "I think @ppizarro was referring to interpreter issues of abruptly interrupting the whole program. " The point I was making is that Go does the same, I was not able to find any kind of special handling for Exit, just a hook for the race detector:

```
	if code == 0 {
		// Give race detector a chance to fail the program.
		// Racy programs do not have the right to finish successfully.
		runtime_beforeExit()
	}
```

If this is a problem in nash we could use panic as a mechanism of exiting, and pass the status code. Nash in some point should catch this panic, recover, get the status code, and them call Exit (after all deferreds have been called, cleanup, etc). It is more convoluted, would not do it that way if strictly necessary.

internal/sh/builtin/exit_linux.go
katcipis: If even with the Atoi thing it is sufficiently portable I rename it with no problem :+1:

internal/sh/builtin/exit_linux_test.go
tiago4orion: Agreed. Maybe only explain that `exit` tests should not `exit` the test infrastructure itself and then this requires executing it on nash sub-processes. The nash version must be the one under test and not the one installed in the system.

internal/sh/builtin/exit_linux_test.go
katcipis: Great :-)

katcipis: Since even when it is plan9 the Go API deliver an int and I have to call Itoa I'm not seeing too much use on getstatus returning a string, right now it returns a int, it seems useless to do these conversions, or am I missing something ?

ppizarro: Maybe we can execute the scripts in another goroutine and the main function blocks waiting for a signal on a channel (done). This channel is passed to goroutine and when the script has finished it send the exit value in this channel.
  
https://github.com/NeowayLabs/nash/blob/master/cmd/nash/main.go#L82

The value readed from channel will be the return value of the main function.
Before return the main function should be "close" everything and stop the goroutine script if it still running (exit gracefully).

The Exit nash function just sent a exit value in this channel and maybe close the channel. 

To do this we must stop the script running in the goroutines. Maybe to each statement executed by the script we can check if the channel was closed.

ppizarro:
Maybe we can execute the scripts in another goroutine and the main function blocks waiting for a signal on a channel (done). This channel is passed to goroutine and when the script has finished it send the exit value in this channel.

https://github.com/NeowayLabs/nash/blob/master/cmd/nash/main.go#L82

The value readed from channel will be the return value of the main function.
Before return the main function should be "close" everything and stop the goroutine script if it still running (exit gracefully).

The Exit nash function just sent a exit value in this channel.

To do this we must stop the script running in the goroutines. An ideia it is the main function closes the channel and to each statement executed by the script we can check if the channel was closed.

tiago4orion:
@ppizarro I don't think it worth the effort...

tiago4orion:
@ppizarro This will be a complicated solution to a problem that even exists yet.

ppizarro:
I totally agree ;)

I just commented do not mean we should do it now ;)

katcipis:
This PR is waiting to add integers on the language, so we can use integers on calls to exit.

katcipis:
Actually we need to remove the "exit not implemented" code and add a plan9 implementation. Later we are going to support exit with integers on a generic manner:

```
exit("0")
exit(0)
```

katcipis:
@tiago4orion While I was looking for a way to implement for plan9 I found out a common interface between the posix and the plan9 implementation. This is not explicit and could break with future Go changes, but the other implementation was also depending on some objects that could change on the future. The current solution does not need different code for each platform and depends only on a single method that is already public today (possibly not going to be removed or changed).

But I'm open to it being a terrible idea :-)

tiago4orion:
Great!!! Much better!

If the implementations change in the future, only our tests would fail. If it happens someday, then we revisit this. 

:-)

katcipis:
@tiago4orion can you test on plan9 to see if works ?

tiago4orion:
yep! But not today =(

katcipis:
sure, after the test I think this can be merged

internal/sh/builtin/exit.go
```diff
@@ -0,0 +1,55 @@
...
+		return errors.NewError(
+			"exit expects a status string, but a %s was provided",
+			obj.Type(),
+		)
+	}
+	statusstr := obj.(*sh.StrObj).Str()
+	status, err := strconv.Atoi(statusstr)
+	if err != nil {
+		return fmt.Errorf(
+			"exit:linux:error[%s] converting status[%s] to int",
```
tiago4orion: this linux in err msg doesn't make sense anymore

katcipis: Woops, fixing it

katcipis: We are good to go :+1:

PR #189, opened 2017-03-10
update dockerfile to go1.8

tiago4orion:
@vitorarins Had you set up automatic build of image in the official docker hub? Or I'll need to push the image manually?


lborguetti:
I think Docker hub build this automatically ...

PR #205, opened 2017-04-15
Remove override of StdErr in pipe commands

### LINKED ISSUE #96: cmd redirection not working properly with pipes
``` sh
# works
json <= -aws ec2 describe-vpcs --vpc-id $vpcId >[2=]

# do not ..
json <= -aws ec2 describe-vpcs --vpc-id $vpcId >[2=] | xargs echo -n
```

vitorarins:
Fixes #96

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -948,7 +948,6 @@ func (shell *Shell) executePipe(pipe *ast.PipeNode) (sh.Obj, error) {
 		}
 
 		cmd.SetStdin(shell.stdin)
-		cmd.SetStderr(shell.stderr)
 
 		if i < last {
 			closeFiles, err = shell.setRedirects(cmd, nodeCmd.Redirects())
@@ -972,8 +971,6 @@ func (shell *Shell) executePipe(pipe *ast.PipeNode) (sh.Obj, error) {
 		var (
 			stdin io.ReadCloser
 		)
-
-		cmd.SetStderr(shell.stderr)
 
 		stdin, err = cmd.StdoutPipe()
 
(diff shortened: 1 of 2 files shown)


vitorarins:
Having a weird race condition with this one, as you can see here: https://travis-ci.org/NeowayLabs/nash/jobs/222288469#L653

tiago4orion:
woww. This races seems to be in the test because the way `testShellExec` collects stdout/stderr.
https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell_test.go#L76

but I'm not sure

vitorarins:
Indeed. I was using the tests of Pipe like they were before, and had the race problem. So I refactored to use `testShellExec` and the problem persisted. I will investigate some more.

tiago4orion:
probably will need to rethink the way of we're collecting stdout and stderr of nash scripts in the test. In the current way we're setting the internal shell.stdout and shell.stderr with empty buffers and this works in simple tests that do not involves output redirection. I'm still not sure, but maybe tests that involves redirection will require spawn the nash binary passing the script to be tested and collects the real stdout/stderr of this process.... 

I made a quick look in the output redirection code and found a problem (but probably not related): In case of redirection `>[2=]` we're not closing the file (probably leaking fds):
https://github.com/NeowayLabs/nash/blob/11c878ebafca5be1cf246c1c173ade74604f95b6/internal/sh/shell.go#L1215
Compare with: https://github.com/NeowayLabs/nash/blob/11c878ebafca5be1cf246c1c173ade74604f95b6/internal/sh/shell.go#L1207

Would be nice if you address this in this PR.

vitorarins:
@tiago4orion I have refactored the test case the way you suggested. Take a look when you can.

vitorarins:
@tiago4orion I reverted the trim space like we discussed and fixed the tests on OSX using `tr` instead of `wc`.

PR #206, opened 2017-04-19
Add printf

katcipis:
Usage on docs

### DIFF
--- a/internal/sh/builtin/append.go
+++ b/internal/sh/builtin/append.go
@@ -22,11 +22,7 @@ func (appendfn *appendFn) ArgNames() []string {
 	return []string{"list"}
 }
 
-func (appendfn *appendFn) Run(
-	stdin io.Reader,
-	stdout io.Writer,
-	stderr io.Writer,
-) ([]sh.Obj, error) {
+func (appendfn *appendFn) Run(in io.Reader, out io.Writer, err io.Writer) ([]sh.Obj, error) {
 	newobj := append(appendfn.obj, appendfn.arg)
 	return []sh.Obj{sh.NewListObj(newobj)}, nil
 }
--- a/internal/sh/builtin/append.go
+++ b/internal/sh/builtin/append.go
@@ -1,6 +1,8 @@
 package builtin
 
 import (
+	"io"
+
 	"github.com/NeowayLabs/nash/errors"
 	"github.com/NeowayLabs/nash/sh"
 )
@@ -20,7 +22,11 @@ func (appendfn *appendFn) ArgNames() []string {
 	return []string{"list"}
 }
 
-func (appendfn *appendFn) Run() ([]sh.Obj, error) {
+func (appendfn *appendFn) Run(
+	stdin io.Reader,
+	stdout io.Writer,
+	stderr io.Writer,
+) ([]sh.Obj, error) {
 	newobj := append(appendfn.obj, appendfn.arg)
 	return []sh.Obj{sh.NewListObj(newobj)}, nil
 }
--- a/internal/sh/builtin/printf.go
+++ b/internal/sh/builtin/printf.go
@@ -20,15 +20,11 @@ func newPrintf() *printfFn {
 }
 
 func (s *printfFn) ArgNames() []string {
-	return []string{"fmt", "args"}
+	return []string{"fmt", "args..."}
 }
 
-func (s *printfFn) Run(
-	stdin io.Reader,
-	stdout io.Writer,
-	stderr io.Writer,
-) ([]sh.Obj, error) {
-	fmt.Fprintf(stdout, s.fmt, s.args...)
+func (s *printfFn) Run(in io.Reader, out io.Writer, err io.Writer) ([]sh.Obj, error) {
+	fmt.Fprintf(out, s.fmt, s.args...)
 	return nil, nil
 }
 
--- a/internal/sh/builtin/printf.go
+++ b/internal/sh/builtin/printf.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/NeowayLabs/nash/errors"
 	"github.com/NeowayLabs/nash/sh"
 )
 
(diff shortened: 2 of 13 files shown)


katcipis:
I was thinking on just calling it "print" instead of "printf" since we don't need both functions.

tiago4orion:
I prefer `printf`. If it's called print I would need to look in the docs to know if it format the arguments or not..
And this way we can create a sprintf function in std nash lib by doing something like `out <= printf(fmt, arg)`.

internal/sh/builtin/printf.go
```diff
@@ -0,0 +1,46 @@
...
+		args []interface{}
+	}
+)
+
+func newPrintf() *printfFn {
+	return &printfFn{}
+}
+
+func (s *printfFn) ArgNames() []string {
+	return []string{"fmt", "args"}
```
tiago4orion: maybe self document args as `args...` because it's variadic.

internal/sh/builtin/append.go
```diff
@@ -20,7 +22,11 @@ func (appendfn *appendFn) ArgNames() []string {
 	return []string{"list"}
 }
 
-func (appendfn *appendFn) Run() ([]sh.Obj, error) {
+func (appendfn *appendFn) Run(
```
tiago4orion: Maybe if you name the variables `in, out, err` you can keep in one line.

katcipis:
@tiago4orion yeah I was thinking about languages like python/lua that usually only have a print and that is it.

I was thinking about your use case, but is that even possible ? If we write:

```
a <= printf("kakaka")
```

How we disambiguate if we are getting the return values of the function or its stdout ? In the end I was thinking on implementing sprintf as a builtin too.

tiago4orion:
Yeah, you're right =) I'd confused with `a <= printf "%s:%s" filename linenum`...
I format some strings already using the command printf this way and forgot the function semantics.
I builtin `sprintf` is ok also, it was only one idea.

Regarding the name `printf` or `print` I like both. We can wait other people's opinion before decide.

internal/sh/builtin/printf_test.go
```diff
@@ -0,0 +1,128 @@
...
+			output: "helloworld",
+		},
+		"fmtstring": {
+			script: `printf("%s:%s", "hello", "world")`,
+			output: "hello:world",
+		},
+		"fmtlist": {
+			script: `
+				list = ("1" "2" "3")
+				printf("%s:%s", "list", $list)
```
tiago4orion: I don't know if this is the right behavior. [sh.ListObj](https://github.com/NeowayLabs/nash/blob/master/sh/obj.go#L69) can stringify itself (every nash type can...), but this is internal (user should know?). The formatting expect a `%s` and we're passing a list. Will we print any object as string? In other words, %s would be useful for any kind of object?
I don't know if this is bad , actually I do think this nice but would like to know your opinion and how we can document this behavior (different from C/Go printf).

I'm thinking in extend the syntax to support Go-style variadic arguments also, then the code below should work:

```
os = ("plan9" "from" "outer" "space")
printf("%s %s %s %s", os...)
```
What do you think?

internal/sh/builtin/printf_test.go
```diff
@@ -0,0 +1,128 @@
...
+		"listasfmt": {
+			script: `
+				list = ("%s" "%s")
+				printf($list, "1", "2")
+			`,
+			output: "1 2",
+		},
+		"invalidFmt": {
+			script: `printf("%d%s", "invalid")`,
+			output: "%!d(string=invalid)%!s(MISSING)",
```
tiago4orion: bleh, Go fmt parser leaking here.. But I don't know of a better way of handling this other than we making the formatting our self... For now, it's ok.

internal/sh/builtin/printf_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	tests := map[string]printfDesc{
+		"noParams": {
+			script: `printf()`,
+		},
+	}
+
+	for name, desc := range tests {
+		t.Run(name, func(t *testing.T) {
```
tiago4orion: I do not understand what you're testing here =(
Apparently you're testing if printf could be used without arguments, but the test name confuses me.

internal/sh/shell.go
```diff
@@ -438,6 +447,14 @@ func (shell *Shell) setupBuiltin() {
 	}
 }
 
+func (shell *Shell) updateBuiltinFnIO() {
```
tiago4orion: :+1:

internal/sh/builtin/printf_test.go
katcipis: Did not liked it either, but I felt compelled to document the behavior :-)

internal/sh/builtin/printf_test.go
katcipis: I wanted to make it clear that I'm testing what happens when you call printf without parameters. What name would you suggest ?

tiago4orion: sorry, my fault.. I'd misinterpreted the test. I thought printf could be used without arguments, but it cannot. Now the test name makes sense.

internal/sh/builtin/append.go
katcipis: good idea :+1:

internal/sh/builtin/printf.go
katcipis: will do

internal/sh/builtin/printf_test.go
katcipis: The variadic expansion seems nice. About the using slice as format... I don't like it a lot, but since we can simply call printf($list) I had to document what happens if you call printf($list, $anothervalue), and what happens is what is on the test (simplest implementation possible). We can change the behavior, and when it is called with more than one arg I check if the first arg is actually a string.

@vitorarins @tiago4orion what do you think ?

vitorarins:
@katcipis @tiago4orion I think `print` is more suitable because we will probably not have two functions (`print`/`printf`).

PR #207, opened 2017-04-22
Add sprint function

katcipis:


katcipis:
format seems good to me. @vitorarins ?

vitorarins:
Agreed @katcipis @tiago4orion. `format` really makes more sense.

PR #211, opened 2017-05-02
Refactor getCommand to set stuff needed by pipe commands.

### LINKED ISSUE #210: interpreter: pipe commands stuck with no output
When I execute this: 
```
find . | fzf -q na | tr -d "\n"
```
the shell gets stuck in the `fzf` command.

vitorarins:
This is kind of shitty for now. But it fixes the problem.
The tests had to be refactored because of race conditions, but I don't know if it's the right solution.
Also, I had to leave the `SetStdout` call in `executeCommand` because `cmd.StdoutPipe()` is already setting Stdout :disappointed:. Any ideas?

Fixes #210

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -989,6 +989,8 @@ func (shell *Shell) executePipe(pipe *ast.PipeNode) (sh.Obj, error) {
 			stdin io.ReadCloser
 		)
 
+		// StdoutPipe complains if Stdout is already set
+		cmd.SetStdout(nil)
 		stdin, err = cmd.StdoutPipe()
 
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/sh/shell.go
```diff
@@ -1323,6 +1323,9 @@ func (shell *Shell) getCommand(c *ast.CommandNode) (sh.Runner, bool, error) {
 		return nil, ignoreError, err
 	}
 
+	cmd.SetStdin(shell.stdin)
+	cmd.SetStderr(shell.stderr)
```
tiago4orion: As I'd said to you in person today, we can set stdout here also, but set it to nil before calling `cmd.StdoutPipe` in `executePipe`.

internal/sh/shell_test.go
```diff
@@ -1182,20 +1182,49 @@ func TestExecuteBindFn(t *testing.T) {
 }
 
 func TestExecutePipe(t *testing.T) {
+	var stderr bytes.Buffer
```
tiago4orion: what's the problem with the table tests?

tiago4orion: ooh, because of the races we discussed in the other PR, right?

vitorarins: Yep! Any ideas?

internal/sh/shell.go
vitorarins: Yes, will do that soon.

PR #212, opened 2017-05-02
Odd behavior for format

katcipis:
Although the problem does not seem to be related to format.


katcipis:
In the end it was a problem on format

PR #213, opened 2017-05-03
Add support for list literals at function call.

### LINKED ISSUE #214: parser: funcall argument parsing bug
When adding tests for #213 I found a parser bug in the funcall arguments.
See below:

```sh
λ> fn func(a) { echo $a }
λ> func("a" "b")
ERROR: Wrong number of arguments for function func. Expected 1 but found 2
```
The comma seems optional, but this is not a feature hehe
I'll address this issue in #213 also.

tiago4orion:
The code below now is valid:

  ids_luns <= append($ids_luns, ($id $lun))

Closes #214 

Signed-off-by: Tiago Natel de Moura <tiago.natel@neoway.com.br>

### DIFF
--- a/internal/sh/shell_test.go
+++ b/internal/sh/shell_test.go
@@ -1047,8 +1047,7 @@ path="AAA"
 	got := string(out.Bytes())
 	expected := "1"
 	if got != expected {
-		t.Errorf("String differs: '%s' != '%s'", got, expected)
-		return
+		t.Fatalf("String differs: '%s' != '%s'", got, expected)
 	}
 
 }
--- a/internal/sh/shell_test.go
+++ b/internal/sh/shell_test.go
@@ -1032,6 +1032,25 @@ path="AAA"
 		return
 	}
 
+	out.Reset()
+	err = shell.Exec("test fn list arg", `
+	ids_luns = ()
+	id = "1"
+	lun = "lunar"
+	ids_luns <= append($ids_luns, ($id $lun))
+	print(len($ids_luns))`)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	got := string(out.Bytes())
+	expected := "1"
+	if got != expected {
+		t.Errorf("String differs: '%s' != '%s'", got, expected)
+		return
+	}
+
 }
 
 func TestFnComposition(t *testing.T) {
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -1143,15 +1143,21 @@ func (p *Parser) parseFnInv(ident scanner.Token, allowSemicolon bool) (ast.Node,
 			} else {
 				goto parseError
 			}
-		} else {
-			goto parseError
 		}
 
-		if p.peek().Type() == token.Comma {
+		it = p.peek()
+		if it.Type() == token.Comma {
 			p.ignore()
 
 			continue
 		}
+
+		if it.Type() == token.RParen {
+			p.next()
+			break
+		}
+
+		goto parseError
 	}
 
 	// semicolon is optional here
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -1104,7 +1104,6 @@ func (p *Parser) parseFnInv(ident scanner.Token, allowSemicolon bool) (ast.Node,
 	n := ast.NewFnInvNode(ident.FileInfo, ident.Value())
 
 	it := p.next()
-
 	if it.Type() != token.LParen {
 		return nil, newParserError(it, p.name, "Invalid token %v. Expected '('", it)
 	}
@@ -1120,6 +1119,12 @@ func (p *Parser) parseFnInv(ident scanner.Token, allowSemicolon bool) (ast.Node,
 			}
 
 			n.AddArg(arg)
+		} else if it.Type() == token.LParen {
+			listArg, err := p.parseList(nil)
+			if err != nil {
+				return nil, err
+			}
+			n.AddArg(listArg)
 		} else if it.Type() == token.RParen {
 			p.next()
 			break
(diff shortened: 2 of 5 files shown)


internal/sh/shell_test.go
```diff
@@ -1032,6 +1032,25 @@ path="AAA"
 		return
 	}
 
+	out.Reset()
```
katcipis: @tiago4orion there is some new functions on the glob MR that may make it easier to run a script and just get all it's output

internal/sh/shell_test.go
```diff
@@ -1032,6 +1032,25 @@ path="AAA"
...
+	err = shell.Exec("test fn list arg", `
+	ids_luns = ()
+	id = "1"
+	lun = "lunar"
+	ids_luns <= append($ids_luns, ($id $lun))
+	print(len($ids_luns))`)
```
katcipis: print, the more resilient builtin function ever :D

internal/sh/shell_test.go
```diff
@@ -1032,6 +1032,25 @@ path="AAA"
...
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	got := string(out.Bytes())
+	expected := "1"
+	if got != expected {
+		t.Errorf("String differs: '%s' != '%s'", got, expected)
```
katcipis: No love for Fatalf ? No need for returning.

parser/parse.go
```diff
@@ -1120,6 +1119,12 @@ func (p *Parser) parseFnInv(ident scanner.Token, allowSemicolon bool) (ast.Node,
 			}
 
 			n.AddArg(arg)
+		} else if it.Type() == token.LParen {
```
katcipis: lol

internal/sh/shell_test.go
tiago4orion: I know, but it's yet unavailable in master. We can refactor all tests to use it soon.

internal/sh/shell_test.go
tiago4orion: lol

internal/sh/shell_test.go
tiago4orion: I have no love for fatalities :-)

tiago4orion: The problem is exiting when sometimes we need defer cleanups. But I can change in this case, no problem.

katcipis: @tiago4orion I tested once this, and checked fatal implementation, it does not call os.Exit and it will execute all your defers. You can use it safelly ;-)

parser/parse.go
```diff
@@ -1285,33 +1295,10 @@ func (p *Parser) parseReturn(retTok scanner.Token) (ast.Node, error) {
 		}
 
 		if tok.Type() == token.LParen {
-			listInfo := tok.FileInfo
```
tiago4orion: We already had a parseList function but found this piece of code parsing lists also... only changed to use the proper helper function.

internal/sh/shell_test.go
katcipis: Coming soon :+1:

parser/parse.go
katcipis: Great =)

internal/sh/shell_test.go
tiago4orion: true, forgot it! Thanks.

katcipis: well the name IS misleading =/

parser/parse_test.go
```diff
@@ -907,6 +909,68 @@ func TestParseFnInv(t *testing.T) {
...
+	expected.Root = ln
+	parserTest("test fn list arg", `a(("1" "2" "3"))`, expected, t, true)
+
+	// test valid funcall syntaxes (do not verify AST)
+	for _, tc := range []string{
+		`func()`,
+		`func(())`, // empty list
+		`func($a)`,
+		`_($a, $b)`,
+		`__((((()))))`, // perfect fit for a nash obfuscating code contest
```
katcipis: uaheuheuaheuaheuah

parser/parse.go
```diff
@@ -1285,33 +1295,10 @@ func (p *Parser) parseReturn(retTok scanner.Token) (ast.Node, error) {
...
-				if err != nil {
-					return nil, err
-				}
-
-				values = append(values, arg)
-			}
-
-			if tok.Type() != token.RParen {
-				return nil, errors.NewUnfinishedListError(p.name, tok)
+			listArg, err := p.parseList(nil)
```
vitorarins: No problems here, but passing `nil` to this function looks kind of weird. I know the function was already like this but I noticed inside the `parseList` function we only check for `nil` and start working with the `p.next()`.
What do you think about  passing `p.next()` here and that way we can remove that nasty `if` check for `nil`?

tiago4orion: The 'if check' is required inside or outside the parseList function. We need to review the parser because in the past it was a handwritten recursive descent equivalent to ll(1), but to solve a nasty bug I'd to pass around a new stored token (but I'm not sure this is really required) then making it somewhat ll(2)... I only tried to fix the thing fast, there was no review at that time. If this is really required, the best way to improve that is adding room for a new lookahead token in the parser struct.

tiago4orion:
@katcipis Thanks. Spec updated!

tiago4orion:
@vitorarins @katcipis @lborguetti Can we merge?

parser/parse.go
vitorarins: Okay then, like I said no problems. We can talk about it later :wink:

PR #216, opened 2017-05-03
Redirection to file should truncate when possible

tiago4orion:
The redirection tests sucks.. I'll improve that

Closes #216

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -1081,7 +1081,6 @@ func (shell *Shell) openRedirectLocation(location ast.Expr) (io.WriteCloser, err
 	)
 
 	locationObj, err := shell.evalExpr(location)
-
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/sh/shell.go
```diff
@@ -1106,7 +1105,7 @@ func (shell *Shell) openRedirectLocation(location ast.Expr) (io.WriteCloser, err
 	}
 
 	if protocol == "" {
-		return os.OpenFile(locationStr, os.O_RDWR|os.O_CREATE, 0644)
+		return os.OpenFile(locationStr, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
```
katcipis: os.Create has different flags ?

internal/sh/shell_test.go
```diff
@@ -546,30 +546,40 @@ func TestExecuteRedirection(t *testing.T) {
...
+		t.Fatal(err)
+	}
 
+	if string(content) != "hello world" {
+		t.Fatalf("File differ: '%s' != '%s'", string(content), "hello world")
+	}
+
+	// Test redirection truncate the file
+	err = shell.Exec("redirect", `
+        echo -n "a" > `+path+`
```
katcipis: could use fmt.Sprintf to generate the code, don't know what is worse although =)

internal/sh/shell.go
tiago4orion: same flags, but different permissions

katcipis: I see =)

tiago4orion:
@katcipis @vitorarins Can we merge this also?

PR #217, opened 2017-05-03
Fix closures bugs

katcipis:
Just reproducing know problems for now

### DIFF
--- a/internal/sh/functions_test.go
+++ b/internal/sh/functions_test.go
@@ -1,0 +1,50 @@
+package sh
+
+import "testing"
+
+func TestFunctionsClosures(t *testing.T) {
+	for _, test := range []execTestCase{
+		{
+			desc: "simpleClosure",
+			execStr: `
+				fn func(a) {
+					fn closure() {
+						print($a)
+					}
+					return $closure
+				}
+
+				x <= func("1")
+				y <= func("2")
+				$x()
+				$y()
+			`,
+			expectedStdout: "12",
+		},
+		//{
+		//desc: "closuresSharingState",
+		//execStr: `
+		//fn func() {
+		//a = ()
+		//fn add(elem) {
+		//a <= append($a, $elem)
+		//}
+		//fn dump() {
+		//print($a)
+		//}
+		//return $add, $dump
+		//}
+
+		//add, dump <= func()
+		//$add("1")
+		//$add("3")
+		//$dump()
+		//`,
+		//expectedStdout: "1 3",
+		//},
+	} {
+		t.Run(test.desc, func(t *testing.T) {
+			testExec(t, test)
+		})
+	}
+}
--- a/internal/sh/functions_test.go
+++ b/internal/sh/functions_test.go
@@ -21,27 +21,27 @@ func TestFunctionsClosures(t *testing.T) {
 			`,
 			expectedStdout: "12",
 		},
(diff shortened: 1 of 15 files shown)


internal/sh/functions_test.go
```diff
@@ -0,0 +1,50 @@
...
+			`,
+			expectedStdout: "12",
+		},
+		{
+			desc: "closuresSharingState",
+			execStr: `
+				fn func() {
+					a = ()
+					fn add(elem) {
+						a <= append($a, $elem)
```
katcipis: @tiago4orion here we have a problem, on a closure when I write this, am I manipulating the parent **a** variable or am I creating a new **a** variable ?

If I'm creating a new one...how do I manipulate the parent **a** ?

If I'm manipulating the parent **a**, how do I create a new one ? :-)

Lua solves this with a **local** keyword to create new locals, and the default is to manipulate the first symbol found on the stack.

tiago4orion: > @tiago4orion here we have a problem, on a closure when I write this, am I manipulating the parent "a" variable or am I creating a new "a" variable ?

Assignments in nash are always implicitly local. Then, in this case you're creating a new `a` variable valid only inside the `add` function (the closure).
But that's a different nash issue, the lack of ways of updating the parent scope. 
Compare the implementations of [sh.Getvar](https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell.go#L319) e [sh.Setvar](https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell.go#L375).
Sure we'll need to introduce a new syntactic ways to handle that, the problem is that we have several options. I think a proposal issue to discuss this new syntax and trade offs is required.

As this isn't a closure issue (the lack of local/global keyword, or explicit syntax for var declaration - like Go/javascript) I think is better to change this test case to focus in our closure problem.

katcipis: Me not having a way to properly use closures is not a closure problem ? The tests serves as examples and I really would like to see a working tested example on how I do manipulate state when using closures (independent if the default behavior is creating a new var or not, seems good to me to document that behavior as a test).

Anyway, today it is broken because it will not create a new variable, it will use the parent one (I'll change the test to reflect this problem).

tiago4orion: saying in other way: nash closures are only read-only. But they are read-only not because we have a bug in closures, but because nash doesn't have a syntax to update the parent scope. Sure we can merge both issues, but I don't know it's good because it will require different design decisions.

katcipis: I understand, I'll focus on the problems then :-).

Right now this:

```
				fn func() {
					a = ()
					fn add(elem) {
						a <= append($a, $elem)
						print($a)
					}
					return $add
				}

				add <= func()
				$add("1")
				$add("3")
				$add("5")
```

Fails like this:

```
    --- FAIL: TestFunctionsClosures/eachCallCreatesNewVar (0.00s)
       shell_test.go:95: [eachCallCreatesNewVar] Stdout differs: 'a:1,a:3,a:5,' != 'a:1,a:1 3,a:1 3 5,'
```

Basically besides having the two closures pointing to same state problem, there is also this one where a new variable **a** is NOT created each time I call the function add =/

fun ! lol

tiago4orion:
Closes #223

PR #220, opened 2017-05-07
Add support for variadic arguments

tiago4orion:
The '...' syntax was added to declare functions that can receive any number of arguments. 

- A ... parameter is received as a list inside the function.

```sh
fn println(args...) {
        for a in $args {
                print($a)
        }
        print("\n")
}
```
- A ... parameter is optional
```sh
println() # Output: \n

fn a(b, c...) {
    print($b, $c...)
}
a("test") # Output: test
a() # Fail, requires at least one argument.
```
- As a special case, the ... expansion is allowed in command arguments to pass a list as multiples arguments.

```sh
# The commands below are equivalent
curl $args
curl $args...
```

This is only a first proposal, feel free to criticize.

I believe now we can change the `print` builtin to just print (no format) and then add the following functions as nash stdlib:

```sh
fn printf(fmt, args...) {
    print(format(fmt, $args...))
}

fn println(args...) {
        for arg in $args {
                print($arg)
        }
        print("\n")
}

# and so on
```

### DIFF
--- a/ast/node_args.go
+++ b/ast/node_args.go
@@ -84,7 +84,11 @@ func (i *IntExpr) IsEqual(other Node) bool {
 	return i.val == o.val
 }
 
-func NewListExpr(info token.FileInfo, values []Expr, variadic bool) *ListExpr {
+func NewListExpr(info token.FileInfo, values []Expr) *ListExpr {
+	return NewListVariadicExpr(info, values, false)
+}
+
+func NewListVariadicExpr(info token.FileInfo, values []Expr, variadic bool) *ListExpr {
 	return &ListExpr{
 		NodeType: NodeListExpr,
 		FileInfo: info,
@@ -171,10 +175,15 @@ func (c *ConcatExpr) IsEqual(other Node) bool {
 }
 
 func NewVarExpr(info token.FileInfo, name string) *VarExpr {
+	return NewVarVariadicExpr(info, name, false)
+}
+
+func NewVarVariadicExpr(info token.FileInfo, name string, isVariadic bool) *VarExpr {
 	return &VarExpr{
-		NodeType: NodeVarExpr,
-		FileInfo: info,
-		Name:     name,
+		NodeType:   NodeVarExpr,
+		FileInfo:   info,
+		Name:       name,
+		IsVariadic: isVariadic,
 	}
 }
 
@@ -192,7 +201,11 @@ func (v *VarExpr) IsEqual(other Node) bool {
 		v.IsVariadic == o.IsVariadic
 }
 
-func NewIndexExpr(info token.FileInfo, va *VarExpr, idx Expr, variadic bool) *IndexExpr {
+func NewIndexExpr(info token.FileInfo, va *VarExpr, idx Expr) *IndexExpr {
+	return NewIndexVariadicExpr(info, va, idx, false)
+}
+
+func NewIndexVariadicExpr(info token.FileInfo, va *VarExpr, idx Expr, variadic bool) *IndexExpr {
 	return &IndexExpr{
 		NodeType: NodeIndexExpr,
 		FileInfo: info,
--- a/ast/node_args.go
+++ b/ast/node_args.go
@@ -118,10 +118,11 @@ func (l *ListExpr) IsEqual(other Node) bool {
 		return false
 	}
 
-	for i := 0; i < len(l.List); i++ {
-		if !l.List[i].IsEqual(o.List[i]) {
-			debug("%v(%s) != %v(%s)", l.List[i], l.List[i].Type(),
-				o.List[i], o.List[i].Type())
+	for i, val := range l.List {
+		oval := o.List[i]
+		if !val.IsEqual(oval) {
+			debug("%v(%s) != %v(%s)", val, val.Type(),
+				oval, oval.Type())
 			return false
 		}
 	}
--- a/ast/node_args.go
+++ b/ast/node_args.go
@@ -84,20 +84,19 @@ func (i *IntExpr) IsEqual(other Node) bool {
 	return i.val == o.val
 }
 
-func NewListExpr(info token.FileInfo, values []Expr) *ListExpr {
+func NewListExpr(info token.FileInfo, values []Expr, variadic bool) *ListExpr {
 	return &ListExpr{
 		NodeType: NodeListExpr,
 		FileInfo: info,
 
-		list: values,
+		List:       values,
+		IsVariadic: variadic,
 	}
 }
 
-func (l *ListExpr) List() []Expr { return l.list }
-
 // PushExpr push an expression to end of the list
 func (l *ListExpr) PushExpr(a Expr) {
-	l.list = append(l.list, a)
+	l.List = append(l.List, a)
 }
 
 func (l *ListExpr) IsEqual(other Node) bool {
@@ -111,14 +110,14 @@ func (l *ListExpr) IsEqual(other Node) bool {
 		return false
 	}
 
-	if len(l.list) != len(o.list) {
+	if len(l.List) != len(o.List) {
 		return false
 	}
 
-	for i := 0; i < len(l.list); i++ {
-		if !l.list[i].IsEqual(o.list[i]) {
-			debug("%v(%s) != %v(%s)", l.list[i], l.list[i].Type(),
-				o.list[i], o.list[i].Type())
+	for i := 0; i < len(l.List); i++ {
+		if !l.List[i].IsEqual(o.List[i]) {
+			debug("%v(%s) != %v(%s)", l.List[i], l.List[i].Type(),
+				o.List[i], o.List[i].Type())
 			return false
 		}
 	}
@@ -175,11 +174,9 @@ func NewVarExpr(info token.FileInfo, name string) *VarExpr {
 	return &VarExpr{
 		NodeType: NodeVarExpr,
 		FileInfo: info,
-		name:     name,
+		Name:     name,
 	}
 }
-
-func (v *VarExpr) Name() string { return v.name }
 
 func (v *VarExpr) IsEqual(other Node) bool {
 	if !v.equal(v, other) {
@@ -187,26 +184,24 @@ func (v *VarExpr) IsEqual(other Node) bool {
 	}
 
 	o, ok := other.(*VarExpr)
-
 	if !ok {
 		return false
 	}
 
-	return v.name == o.name
+	return v.Name == o.Name &&
+		v.IsVariadic == o.IsVariadic
 }
 
-func NewIndexExpr(info token.FileInfo, variable *VarExpr, index Expr) *IndexExpr {
+func NewIndexExpr(info token.FileInfo, va *VarExpr, idx Expr, variadic bool) *IndexExpr {
 	return &IndexExpr{
 		NodeType: NodeIndexExpr,
 		FileInfo: info,
 
-		variable: variable,
-		index:    index,
+		Var:        va,
+		Index:      idx,
+		IsVariadic: variadic,
 	}
 }
-
-func (i *IndexExpr) Var() *VarExpr { return i.variable }
-func (i *IndexExpr) Index() Expr   { return i.index }
 
 func (i *IndexExpr) IsEqual(other Node) bool {
 	if !i.equal(i, other) {
@@ -214,10 +209,11 @@ func (i *IndexExpr) IsEqual(other Node) bool {
 	}
 
 	o, ok := other.(*IndexExpr)
-
 	if !ok {
 		return false
 	}
 
-	return i.variable.IsEqual(o.variable) && i.index.IsEqual(o.index)
+	return i.Var.IsEqual(o.Var) &&
+		i.Index.IsEqual(o.Index) &&
+		i.IsVariadic == o.IsVariadic
 }
--- a/ast/node_fmt.go
+++ b/ast/node_fmt.go
@@ -18,16 +18,16 @@ func (i *IntExpr) String() string {
 }
 
 func (l *ListExpr) string() (string, bool) {
-	elems := make([]string, len(l.list))
+	elems := make([]string, len(l.List))
 	columnCount := 0
 	forceMulti := false
 
-	for i := 0; i < len(l.list); i++ {
-		if l.list[i].Type() == NodeListExpr {
+	for i := 0; i < len(l.List); i++ {
+		if l.List[i].Type() == NodeListExpr {
 			forceMulti = true
 		}
 
-		elems[i] = l.list[i].String()
+		elems[i] = l.List[i].String()
 		columnCount += len(elems[i])
 	}
 
@@ -59,11 +59,18 @@ func (c *ConcatExpr) String() string {
 }
 
 func (v *VarExpr) String() string {
-	return v.name
+	if v.IsVariadic {
+		return v.Name + "..."
+	}
+	return v.Name
 }
 
 func (i *IndexExpr) String() string {
-	return i.variable.String() + "[" + i.index.String() + "]"
+	ret := i.Var.String() + "[" + i.Index.String() + "]"
+	if i.IsVariadic {
+		return ret + "..."
+	}
+	return ret
 }
 
 func (l *BlockNode) adjustGroupAssign(node assignable, nodes []Node) {
@@ -564,8 +571,7 @@ func (n *FnDeclNode) String() string {
 	}
 
 	for i := 0; i < len(n.args); i++ {
-		fnStr += n.args[i]
-
+		fnStr += n.args[i].String()
 		if i < (len(n.args) - 1) {
 			fnStr += ", "
 		}
--- a/ast/node_fmt.go
+++ b/ast/node_fmt.go
@@ -596,6 +596,14 @@ func (n *FnDeclNode) String() string {
 	return fnStr
 }
 
+func (arg *FnArgNode) String() string {
+	ret := arg.Name
+	if arg.IsVariadic {
+		ret += "..."
+	}
+	return ret
+}
+
 // String returns the string representation of function invocation
 func (n *FnInvNode) string() (string, bool) {
 	fnInvStr := n.name + "("
--- a/ast/node_fmt.go
+++ b/ast/node_fmt.go
@@ -1,6 +1,7 @@
 package ast
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 )
@@ -66,7 +67,7 @@ func (v *VarExpr) String() string {
 }
 
 func (i *IndexExpr) String() string {
-	ret := i.Var.String() + "[" + i.Index.String() + "]"
+	ret := fmt.Sprintf("%s[%s]", i.Var, i.Index)
 	if i.IsVariadic {
 		return ret + "..."
 	}
--- a/internal/sh/builtin/append.go
+++ b/internal/sh/builtin/append.go
@@ -32,12 +32,12 @@ func (appendfn *appendFn) Run(in io.Reader, out io.Writer, err io.Writer) ([]sh.
 
 func (appendfn *appendFn) SetArgs(args []sh.Obj) error {
 	if len(args) < 2 {
-		return errors.NewError("appendfn expects at least two arguments")
+		return errors.NewError("append expects at least two arguments")
 	}
 
 	obj := args[0]
 	if obj.Type() != sh.ListType {
-		return errors.NewError("appendfn expects a list as first argument, but a %s was provided",
+		return errors.NewError("append expects a list as first argument, but a %s was provided",
 			obj.Type())
 	}
 
--- a/internal/sh/builtin/append.go
+++ b/internal/sh/builtin/append.go
@@ -9,8 +9,8 @@ import (
 
 type (
 	appendFn struct {
-		obj []sh.Obj
-		arg sh.Obj
+		obj  []sh.Obj
+		args []sh.Obj
 	}
 )
 
@@ -18,31 +18,32 @@ func newAppend() *appendFn {
 	return &appendFn{}
 }
 
-func (appendfn *appendFn) ArgNames() []string {
-	return []string{"list"}
+func (appendfn *appendFn) ArgNames() []sh.FnArg {
+	return []sh.FnArg{
+		sh.NewFnArg("list", false),
+		sh.NewFnArg("value...", true), // variadic
+	}
 }
 
 func (appendfn *appendFn) Run(in io.Reader, out io.Writer, err io.Writer) ([]sh.Obj, error) {
-	newobj := append(appendfn.obj, appendfn.arg)
+	newobj := append(appendfn.obj, appendfn.args...)
 	return []sh.Obj{sh.NewListObj(newobj)}, nil
 }
 
 func (appendfn *appendFn) SetArgs(args []sh.Obj) error {
-	if len(args) != 2 {
-		return errors.NewError("appendfn expects two arguments")
+	if len(args) < 2 {
+		return errors.NewError("appendfn expects at least two arguments")
 	}
 
 	obj := args[0]
-
 	if obj.Type() != sh.ListType {
 		return errors.NewError("appendfn expects a list as first argument, but a %s[%s] was provided", obj, obj.Type())
 	}
 
-	arg := args[1]
-
+	values := args[1:]
 	if objlist, ok := obj.(*sh.ListObj); ok {
 		appendfn.obj = objlist.List()
-		appendfn.arg = arg
+		appendfn.args = values
 		return nil
 	}
 
--- a/internal/sh/builtin/append.go
+++ b/internal/sh/builtin/append.go
@@ -37,7 +37,8 @@ func (appendfn *appendFn) SetArgs(args []sh.Obj) error {
 
 	obj := args[0]
 	if obj.Type() != sh.ListType {
-		return errors.NewError("appendfn expects a list as first argument, but a %s[%s] was provided", obj, obj.Type())
+		return errors.NewError("appendfn expects a list as first argument, but a %s was provided",
+			obj.Type())
 	}
 
 	values := args[1:]
--- a/internal/sh/builtin/append_test.go
+++ b/internal/sh/builtin/append_test.go
@@ -52,14 +52,14 @@ func testAppend(t *testing.T, tc testcase) {
 func TestAppend(t *testing.T) {
 	for _, tc := range []testcase{
 		{
-			name:        "at least two arguments",
+			name:        "no argument fails",
 			code:        `append()`,
-			expectedErr: "<interactive>:1:0: appendfn expects at least two arguments",
+			expectedErr: "<interactive>:1:0: append expects at least two arguments",
 		},
 		{
-			name:        "at least two arguments",
+			name:        "one argument fails",
 			code:        `append("1")`,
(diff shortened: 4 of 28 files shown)


katcipis:
Awesome @tiago4orion :-)

Even when I was developing print as a builtin I was not able to see an advantage on having print and printf. Do you see one ?

Just calling print("msg") already works, why have another function ?

The only advantage seems to be simpler builtin print, but it will be marginally simpler.

Building println on top of print is indeed a great idea, just need a good stdlib to add it on :+1:

tiago4orion:
I think some way to print without formatting is useful:
For example, how we can print a raw string (eg.: string having 'format chars') ?
```sh
a = "{\"error\": \"%s\"}"
print($a) # Output: {"error": "%!s(MISSING)"}
```
For debugging purposes, it's important that we have some way to print stuff as is.
We can introduce another built-in to solve that, maybe call it `puts` (that just prints, no format and no newline), but then I would argue that `print` should be implemented in nash using this low level print:

```
fn print(fmt, args...) {
    puts(format(fmt, args...))
}
```

What do you think?

katcipis:
@tiago4orion I see, I think I never needed this kind of stuff, but indeed it is possible. Now it makes sense :-)

katcipis:
I'm not sure if I'm just tripping here, but the case:

```
curl $args
curl $args...
```

Made me think on a way to make the language more clear/explicit. I always found it odd to rely on the string representation of a list of strings to enable the user to pass a list of parameters for commands. I always felt that a change on the strings representation of a type would break another very important feature of the language.

Also, I always got confused why sometimes cmd $arg not worked, because it does not work when the list has other lists inside.

Enabling:

```
cmd $args...
```
As the only valid way to unpack a list of arguments seems much more clear to me, changing the strings representation of a list would not break anything, because the expansion of the list as arguments rely on a feature that is made explicitly to do that.

In conclusion, my proposal is to deprecate **cmd $arg** when $arg is a list. Commands ONLY works with string arguments, it will work with the expansion because it will expand the list to multiple strings args, on the case that it expands to something else, it will break, and it will be expected to break.

If we dont do this, lets say I have this:

```
a = ("a" ("b"))
echo $a...
```

It seems like it will work, because it will expand, and the list with strings inside are also allowed on commands, but it seems very confusing to me the possible combinations and behaviors.

Feel free to consider this just a limitation of my small skull :-)

ast/node_args.go
```diff
@@ -111,14 +114,14 @@ func (l *ListExpr) IsEqual(other Node) bool {
...
-	if len(l.list) != len(o.list) {
+	if len(l.List) != len(o.List) {
 		return false
 	}
 
-	for i := 0; i < len(l.list); i++ {
-		if !l.list[i].IsEqual(o.list[i]) {
-			debug("%v(%s) != %v(%s)", l.list[i], l.list[i].Type(),
-				o.list[i], o.list[i].Type())
+	for i := 0; i < len(l.List); i++ {
```
katcipis: I would avoid all those arrays accesses by using a for i, val and also creating anther variable to contain o.List[i]. The arrays accesses are copied everywhere, it is a little harder to read and easier to make a mistake (not a big thing although).

ast/node_fmt.go
```diff
@@ -59,11 +59,18 @@ func (c *ConcatExpr) String() string {
...
-	return v.name
+	if v.IsVariadic {
+		return v.Name + "..."
+	}
+	return v.Name
 }
 
 func (i *IndexExpr) String() string {
-	return i.variable.String() + "[" + i.index.String() + "]"
+	ret := i.Var.String() + "[" + i.Index.String() + "]"
```
katcipis: If you use:

```
ret := fmt.Sprintf("%s[%s]", i.Var, i.Index)
```
Won't Go automatically call the method String from your type ? (not sure)

It would be easier to read, and less stuff to write :-)

internal/sh/fncall.go
```diff
@@ -41,22 +46,79 @@ func NewUserFn(name string, parent *Shell) (*UserFn, error) {
...
 func (fn *UserFn) SetArgs(args []sh.Obj) error {
-	if len(fn.argNames) != len(args) {
-		return errors.NewError("Wrong number of arguments for function %s. Expected %d but found %d",
+	var (
+		isVariadic      bool
+		countNormalArgs int
+	)
+
+	for i := 0; i < len(fn.argNames); i++ {
+		argName := fn.argNames[i]
```
katcipis: Why not:
```
for i, argName := 0; i < len(fn.argNames); i++ {
```

Not C enough ? =)

internal/sh/fncall.go
```diff
@@ -41,22 +46,79 @@ func NewUserFn(name string, parent *Shell) (*UserFn, error) {
...
+			// there's only a variadic (optional) argument
+			// and user supplied no argument...
+			// then only initialize the variadic variable to
+			// empty list
+			fn.Setvar(fn.argNames[0].Name, sh.NewListObj([]sh.Obj{}))
+			return nil
+		}
+	}
+
+	var i int
```
katcipis: To a variable that is even used outside the loop scope **i** seems to be a too short name, broader scope demands improved naming

tiago4orion:
I'm not sure I understand what you mean here:
> I always found it odd to rely on the string representation of a list of strings to enable the user to pass a list of parameters for commands. I always felt that a change on the strings representation of a type would break another very important feature of the language.

When you pass a list to a command, as in `cmd $arg`, it does not rely in `arg.String()`, but it proper iterate the list and check if it's composed of strings. The code is here:
https://github.com/NeowayLabs/nash/blob/master/internal/sh/cmd.go#L79

> Also, I always got confused why sometimes cmd $arg not worked, because it does not work when the list has other lists inside.

Well, in every operating system (that I know) arguments are just strings (or null-terminated byte array in unix-like OS). That's why nash aborts if the list isn't a list of strings.. not a nash limitation of the string representation of a list. But maybe I misunderstand your assumption.

internal/sh/shell.go
```diff
@@ -1476,17 +1481,57 @@ func (shell *Shell) evalIndexedVar(indexVar *ast.IndexExpr) (sh.Obj, error) {
...
+		return nil, errors.NewEvalError(shell.filename, indexVar.Var,
+			"Invalid indexing of non-list variable: %s (%+v)", v.Type(), v)
+	}
+
+	indexNum, err := shell.evalIndex(indexVar.Index)
+	if err != nil {
+		return nil, err
+	}
+
+	vlist := v.(*sh.ListObj)
```
katcipis: Allowing panic here is desired ? (I imagine that in this case it is, just to be sure)

internal/sh/shell.go
```diff
@@ -1499,22 +1544,55 @@ func (shell *Shell) evalVariable(a ast.Expr) (sh.Obj, error) {
 	}
 
 	vexpr := a.(*ast.VarExpr)
-	varName := vexpr.Name()
+	varName := vexpr.Name
 
-	if v, ok = shell.Getvar(varName[1:]); !ok {
+	if value, ok = shell.Getvar(varName[1:]); !ok {
```
katcipis: are there guarantees that the varName will always have a size > 1 ?

internal/sh/shell.go
```diff
@@ -1499,22 +1544,55 @@ func (shell *Shell) evalVariable(a ast.Expr) (sh.Obj, error) {
 	}
 
 	vexpr := a.(*ast.VarExpr)
```
katcipis: Another panic. I understand that these are very exceptional situations, but I'm afraid that if they happen the panic message will not be very clear on what went wrong (although the stack usually is enough to track down, if the person knows the code well). I honestly do not like the idea of the language tossing up panics from Go. It gives more work but it seems nicer to provide nice nash errors, even for fatal stuff (seems like we are leaking Go if something goes wrong on our code).

internal/sh/shell.go
```diff
@@ -1525,6 +1603,79 @@ func (shell *Shell) evalExprs(exprs []ast.Expr) ([]sh.Obj, error) {
...
+		}
+
+		ret = append(ret, objs...)
+	}
+
+	return ret, nil
+}
+
+func (shell *Shell) evalArgExpr(expr ast.Expr) ([]sh.Obj, error) {
+	switch expr.Type() {
```
katcipis: Instead of having this switch case would it not be better if each ast.Expr had a Eval method (interface level). Besides removing the switch you would also remove the casts (which are also not very desirable).

Doing like that, every type concentrates the code to handle that type, the way it is done today handling of each type is spread through the code on different switch cases (perhaps on this case is just this one, but they tend to multiply).

internal/sh/shell.go
```diff
@@ -1525,6 +1603,79 @@ func (shell *Shell) evalExprs(exprs []ast.Expr) ([]sh.Obj, error) {
...
+			return []sh.Obj{objs[0]}, nil
+		}
+	}
+
+	return nil, errors.NewEvalError(shell.filename,
+		expr, "Failed to eval expression: %+v", expr)
+}
+
 func (shell *Shell) evalExpr(expr ast.Expr) (sh.Obj, error) {
 	switch expr.Type() {
```
katcipis: Yeah, and the switch case multiples =)

katcipis:
Sorry, once we talked about that and I understood that the string representation of a list of strings matches with the arguments required for the command (printing a list of strings generates a space separated sequence of strings, matches with what the command expects), and that is why it worked. Even tough, I'm not sure if it good to have these two ways, it made sense when we did not have vargs and expansion, now that we have it seems a little confusing to me, but it is not a big deal.

internal/sh/shell_test.go
```diff
@@ -1216,7 +1216,7 @@ func TestExecutePipe(t *testing.T) {
 	}
 
 	expectedOutput := "hello"
-	actualOutput   := string(stdout.Bytes())
+	actualOutput := string(stdout.Bytes())
```
katcipis: bytes.Buffer already has a String method.

katcipis: For this kinda of e2e testing it seems practical to create a small package that execs a command and already collects output in memory, like Go's Cmd Output() and CombinedOuptut()

tiago4orion:
Yeah, I'm inclined to deprecate the old way also, but today this feature is used everywhere... a big break change.
We can deprecate it now, and remove when we launch the 1.0

internal/sh/shell_test.go
```diff
@@ -2301,3 +2301,73 @@ world`)
...
+		},
+		{
+			desc: "the first argument isn't optional",
+			execStr: `fn a(b, c...) {
+    print($b, $c...)
+}
+a()`,
+			expectedErr: "Wrong number of arguments for function a. Expected at least 1 arguments but found 0",
+		},
+	} {
```
katcipis: It seems to be missing a test for the case of a empty list being expanded (I'm curious on the behavior =)).

parser/parse.go
```diff
@@ -777,7 +792,13 @@ func (p *Parser) parseList(tok *scanner.Token) (ast.Node, error) {
 	}
 
 	p.ignore()
-	return ast.NewListExpr(lit.FileInfo, values), nil
+
+	var isVariadic bool
+	if p.peek().Type() == token.Dotdotdot {
```
katcipis: auheuhaeuhuaheuaheuaheuahe

dotdotdot =)

parser/parse.go
```diff
@@ -1017,24 +1038,27 @@ func (p *Parser) parseFnArgs() ([]string, error) {
 
 	for {
```
katcipis: Living dangerously lol

parser/parse.go
```diff
@@ -1388,7 +1406,7 @@ func (p *Parser) parseFor(it scanner.Token) (ast.Node, error) {
 	if (it.Type() == token.Ident || it.Type() == token.Variable) && next.Type() == token.LParen {
```
katcipis: Smells like a switch case disguised as ifs =)

ast/node_args.go
```diff
@@ -85,19 +85,22 @@ func (i *IntExpr) IsEqual(other Node) bool {
...
 		NodeType: NodeListExpr,
 		FileInfo: info,
 
-		list: values,
+		List:       values,
+		IsVariadic: variadic,
 	}
 }
 
-func (l *ListExpr) List() []Expr { return l.list }
```
tiago4orion: I'm following your suggestion and removing the getters whenever possible =)

ast/node_args.go
tiago4orion: :+1:

spec.ebnf
```diff
@@ -59,7 +59,7 @@ returnDecl = "return" [ ( variable | stringLit | list | fnInv ) ] .
 fnInv = ( variable | identifier ) "(" fnArgValues ")" .
 
 fnArgValues = { fnArgValue [ "," ] } .
-fnArgValue  = [ stringLit | stringConcat | list | variable | fnInv ] .
+fnArgValue  = [ stringLit | stringConcat | list | (variable [ "..." ]) | (list [ "..." ]) fnInv ] .
```
katcipis: I think there is a way to build a BNF that expresses that the variadic arg is only allowed as the last param (not completely sure, it is late, but if you agree I can give it a shot =))

ast/node_fmt.go
tiago4orion: yes, it will call our String() method. I can change, but sometimes I'm worried about the complexity of fmt package =/ But I've brain damage kkk

katcipis: @tiago4orion take it easy on all this anxiety, it will give you headaches =P

internal/sh/builtin/append.go
```diff
@@ -9,40 +9,41 @@ import (
...
+func (appendfn *appendFn) ArgNames() []sh.FnArg {
+	return []sh.FnArg{
+		sh.NewFnArg("list", false),
+		sh.NewFnArg("value...", true), // variadic
+	}
 }
 
 func (appendfn *appendFn) Run(in io.Reader, out io.Writer, err io.Writer) ([]sh.Obj, error) {
-	newobj := append(appendfn.obj, appendfn.arg)
+	newobj := append(appendfn.obj, appendfn.args...)
```
tiago4orion: forgot to document and test this change :disappointed: 
Now `append` builtin supports multiple arguments.

katcipis: I noticed that you changed that on the beginning of the review, but got lost on other stuff and forgot to mention that I did not found the test anywhere =P

internal/sh/fncall.go
tiago4orion: o.O

katcipis: Sorry, I gave a bugged example of what I had in mind =P

internal/sh/shell.go
tiago4orion: Few lines above we check the type of `v`, then here it should be a ListObj. If it's not the case, then the shell is at an inconsistent state, right? This seems to be a good candidate to a panic because it's not safe to continue (a bug was triggered or a race happened somewhere). What do you think? And if is better to validate here, what should I do?

internal/sh/shell.go
tiago4orion: A variable always start with a $ sign

internal/sh/shell.go
katcipis: I understand that it is not safe to continue, but usually errors that you produce directly on the code will give a better hint of what can be wrong than a generic Go panic of invalid casting. But if in this case you can't come up with anything better it is ok =)

internal/sh/shell.go
katcipis: This array has been validated for this before on another part of the code ? That part of the code is completely bug/bullet proof ? If it is not, which kind of error will explode ? Will it be easy to understand/debug ?

internal/sh/fncall.go
tiago4orion:
```
for i, argName := range fn.argNames {}
```
?
:joy: 
what's the problem with `initializations; cond; increment` ? too much typing?
I can change, but I use this a lot..

internal/sh/fncall.go
tiago4orion: It's not used in a broader scope.. It's used in the function scope and the meaning is the same: indexing

From the `PikeStyle`:
_Indices are just notation, so treat them as such._

In this case I cannot think of a better name..

internal/sh/shell.go
tiago4orion: I don't think so.. The interpreter is just eval of data structures created by parser and every node of AST has a proper type information for we rely upon.. otherwise every eval will be a mess of _if else logic_ to catch bugs in previous stages of the interpreter. In this stage (the evaluator) the only thing we can say is that: _parser or scanner has a bug_. And I think this is useless as a stacktrace to the user. For we (maintainers) the stack is much more useful.

The issue is that our `ast` package is exported, then users can generate ASTs and feed the interpreter directly (code generation, and so on) and then this way the evaluator will trigger panics if something do not conform.
To solve that we can work in the `ast` package to add more guarantees, validating nodes created by users or parser, then this panic will never be triggered here, maybe this will require changing the direct access to the struct members to setters and getters again...

Validating nodes in the evaluator will confuse the logic and compromise readability (I think), but I'm open to suggestions.

internal/sh/shell.go
tiago4orion: The scanner only emits `token.Variable` when it found a '$' followed by an identifier:
https://github.com/NeowayLabs/nash/blob/master/scanner/lex.go#L276

The again this code will fail only if the `scanner` is buggy generating `token.Variable` for wrong tokens.

tiago4orion: We can take it simple or make it terrible verbose and spaguetti...
Better to guarantee mistakes by adding tests than by adding never reached lines of code.

internal/sh/shell.go
tiago4orion: you're arguing to mix the interpreter with the AST? I don't know if this is a good design decision...
To eval a node we need the environment (the set of variables, functions, bindings and scopes created) of the shell, then the node will need the shell, but the shell need to import the nodes to eval... There's cyclic dependency here apart from being bad giving eval powers to each node type.

tiago4orion: > Doing like that, every type concentrates the code to handle that type, the way it is done today handling of each type is spread through the code on different switch cases (perhaps on this case is just this one, but they tend to multiply).

Before this PR, nash had only one `evalExpr` and then this switch case occurred in only one place (one function).
But now with variadic arguments I need to evaluate a expression a bit different if it's in a function argument.. Because '...' cannot be used in other places.
My first try was adding a new boolean flag to `evalExprs` and making it handle both cases.. but the code looked messy and then replicated a bit of the logic to make the code more explicit of what it is being doing.

internal/sh/shell.go
```diff
@@ -1884,42 +2033,13 @@ func (shell *Shell) executeIfNotEqual(n *ast.IfNode) ([]sh.Obj, error) {
 	return nil, nil
 }
 
-func (shell *Shell) executeFn(fn sh.Fn, nodeArgs []ast.Expr) ([]sh.Obj, error) {
```
tiago4orion: found an unused function... o.O

internal/sh/shell_test.go
tiago4orion: > bytes.Buffer already has a String method.

Thanks

vitorarins: Yeah, I was thinking about this. In my last PR I didn't want to add another `execTestCase...` function just for this. But I think the idea of creating a package is great. :beers:

internal/sh/builtin/append.go
tiago4orion: :+1:

internal/sh/shell_test.go
tiago4orion: The pkg will need to collect stdout and stderr separately. We're missing to test the stderr of several tests.. looking only in the stdout is not good.. if something not expected are going to stderr we're not tracking.

tiago4orion: I think @katcipis started such `cmd` in their last PR, no?

internal/sh/shell_test.go
tiago4orion: true
:+1:

parser/parse.go
tiago4orion: hey, I did not touch this line =) kkkk
But the boolean expression is: (A || B) && C 
I'm not seeing a switch here, but I must be blindly wrong.

spec.ebnf
tiago4orion: I'm fine. Because of this assumption I validated in the interpreter also with the vararg is the last argument. But making it explicit in the BNF, then we can validate this only in the parser.

parser/parse.go
tiago4orion: It escalated from "Getting dot-dot right" =P
http://plan9.bell-labs.com/sys/doc/lexnames.html

internal/sh/fncall.go
katcipis: No problem at all, it is only odd when you do this:

```
for i := 0; i < len(fn.argNames); i++ {
 		argName := fn.argNames[i]
```

The first thing you do is to create a var to hold the value of fn.argNames[i]. It is smaller and more efficient to just write:

```
for i, argName :=  range fn.argNames{
}
```

I think this was the purpose of the range idiom for arrays/slices =). But to be honest I was more messing around, there is no problem at all with using the increment stuff =P

internal/sh/fncall.go
katcipis: Broader scope = outside the scope of the for loop, not necessarily a package scope =P. But if in this case it is just an index, that is ok, it just sounded a little odd

internal/sh/shell.go
katcipis: Well, thinking that error handling compromises readability is the origin of exceptions =P. If the programming error is really severely stupid perhaps a panic is better, I don't know nash design well enough, sorry

internal/sh/shell.go
katcipis: Did not know it was validated before, just to confirm =)

internal/sh/shell.go
katcipis: Did not noticed that I would mix the interpreter and the AST, sorry =). Would have to get more acquainted with the code/design to make a better proposal. Sorry for the noise

internal/sh/shell.go
katcipis: I use gounused on some projects and it works very well

internal/sh/shell_test.go
katcipis: @tiago4orion yep I did =). I don't get the stderr yet (don't used it)

parser/parse.go
katcipis: I was looking at the "else ifs". It seems to be writable as a switch, and I usuall prefer switches than if elses =P.

All nash code has been written by you =P

spec.ebnf
katcipis: Yeah I noticed the checks for the last param, I was curious on how would be the BNF =)

internal/sh/shell.go
tiago4orion: there's no `error handling` here.. because there's no `error`.
If we substitute this to use exceptions (if go supported them) then still no error will be handled... the only thing we can do here is abort the shell...

But.. If you think stopping using the Type() method of the node and instead use only golanf type assertions and reflection to make the type switchs is a good use case.. then I must agree that we *do* need to treat each type assertion (because only the runtime has the truth.. an internal `type` member).

katcipis: If there is no error then it is ok to not handle errors =), just brainstorming

tiago4orion: no problem bro. Good to discuss that =)

tiago4orion:
This PR introduced a bug when parsing the execution of commands in the current directory:
```sh
λ> ./rkt
ERROR: <stdin line 13>:2:0: Unexpected symbol 'EOF'
```
The fix was:
https://github.com/NeowayLabs/nash/pull/220/commits/f66ec2c71f33445f979a0810fabb99dad4d9f3c9#diff-32d9b0cbc79bb412d52c5ebf6cd1673eR367

I don't know if you guys know it, but nash lexer emits automatic semicolons for us. Then, the code below:

```sh
fn print(arg) {
    echo $arg
}

_, status <= stat -d /etc
if $status == "0" {
    print("/etc exists")
}
```
The parser sees the code below:
```sh
fn print(arg) {
    echo $arg;
}

_, status <= stat -d /etc;
if $status == "0" {
    print("/etc exists");
}
```
The problem was: When scanning the '...', if only one or two dots are found, the lexer emits a token.Arg containing ".." but I forgot to handle the semicolon insertion in this case and then the parser failed because the parser is *very* strict regarding semicolons (like C or Java).
There must be a way to avoid handling token.Arg in two places, but I was not able to get a simple way to do that for now =(

katcipis:
Oops, missed the bug that has been found, sorry

tiago4orion:
relax, impossible to catch all my mistakes in one review =)

tiago4orion:
Added some missing tests and cosmetic changes.
/cc @vitorarins @katcipis

internal/sh/builtin/append.go
```diff
@@ -37,7 +37,8 @@ func (appendfn *appendFn) SetArgs(args []sh.Obj) error {
 
 	obj := args[0]
 	if obj.Type() != sh.ListType {
-		return errors.NewError("appendfn expects a list as first argument, but a %s[%s] was provided", obj, obj.Type())
+		return errors.NewError("appendfn expects a list as first argument, but a %s was provided",
```
katcipis: Wouldn't be better to call "append" instead of "appendfn" ? appendfn does not exist on nash, it may confuse the user.

internal/sh/builtin/append_test.go
```diff
@@ -7,45 +7,106 @@ import (
...
-		return
+func TestAppend(t *testing.T) {
+	for _, tc := range []testcase{
+		{
+			name:        "at least two arguments",
+			code:        `append()`,
+			expectedErr: "<interactive>:1:0: appendfn expects at least two arguments",
+		},
+		{
+			name:        "at least two arguments",
```
katcipis: repeteated name ? Not sure it is a good idea. Perhaps "no arguments fails" and "one argument fails'

ast/node_args.go
```diff
@@ -118,10 +118,11 @@ func (l *ListExpr) IsEqual(other Node) bool {
 		return false
 	}
 
-	for i := 0; i < len(l.List); i++ {
-		if !l.List[i].IsEqual(o.List[i]) {
-			debug("%v(%s) != %v(%s)", l.List[i], l.List[i].Type(),
-				o.List[i], o.List[i].Type())
+	for i, val := range l.List {
+		oval := o.List[i]
```
katcipis: :+1:

ast/node_fmt.go
```diff
@@ -66,7 +67,7 @@ func (v *VarExpr) String() string {
 }
 
 func (i *IndexExpr) String() string {
-	ret := i.Var.String() + "[" + i.Index.String() + "]"
+	ret := fmt.Sprintf("%s[%s]", i.Var, i.Index)
```
katcipis: :+1:

internal/sh/fncall.go
```diff
@@ -58,8 +58,7 @@ func (fn *UserFn) SetArgs(args []sh.Obj) error {
 		countNormalArgs int
 	)
 
-	for i := 0; i < len(fn.argNames); i++ {
-		argName := fn.argNames[i]
+	for i, argName := range fn.argNames {
```
katcipis: :+1:

PR #221, opened 2017-05-11
Add goreleaser to easy the release process

tiago4orion:
Am I missing something?

Bonus of using it:

- Very easy configurable release
- Cross build on every arch/os that Go supports
- Integrates with Travis-CI (auto-generate releases)

Problems I found with goreleaser:

- There's no way to configure different ldflags to each $GOOS
- There's no way to pass build flags to the compiler
- There's no way to release multiple set of binaries (like nash and nashfmt)

Given the limitation of goreleaser, this configuration has the following impacts:

- Nash binary will not be static when building with Go < 1.5
- Nash binary will use the host net resolver instead of netgo when building with Go < 1.5
- Nashfmt isn't shipped in the release...

What do you guys think? Is a good move?

### DIFF
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -1,0 +1,49 @@
+# goreleaser.yml
+build:
+  # Path to main.go file or main package.
+  # Default is `.`
+  main: ./cmd/nash/main.go
+
+  # Name of the binary.
+  # Default is the name of the project directory.
+  binary: nash
+
+  # Custom ldflags template.
+  # This is parsed with Golang template engine and the following variables
+  # are available:
+  # - Date
+  # - Commit
+  # - Tag
+  # - Version (Tag with the `v` prefix stripped)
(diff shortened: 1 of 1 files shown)


vitorarins:
`Given the limitation of goreleaser, this configuration has the following impacts`

These limitations you mentioned are valid only for the binaries generated by goreleaser.
People can still clone the project and build it normally.

That said, I think is a good move. Maybe we can contribute to the goreleaser project and try to ship `nashfmt` with it.

katcipis:
@tiago4orion you said that there is no ldflags to each GOOS, but we still can pass the ldflag "-X main.VersionString" right ?

I don't see a problem with not installing the fmt tool, I don't see a way to avoid some form of bootstraping when nash runs to install the stdlib too =/ (or embed the scripts on the binary ? =P)

PR #228, opened 2017-06-30
Add stdlib

katcipis:
Work in progress


Makefile
```diff
@@ -1,8 +1,32 @@
...
+	go build $(buildargs) -o ./cmd/nash/nash ./cmd/nash
+	go build $(buildargs) -o ./cmd/nashfmt/nashfmt ./cmd/nashfmt
+
+guard-%:
+	@ if [ "${${*}}" = "" ]; then \
+                echo "'$*' must be set in order to install and use nash"; \
+                exit 1; \
+        fi
+
+install: build guard-NASHPATH
```
tiago4orion: I'm not sure if this is a good idea. The nash binary already looks in a default NASHPATH location if it's set and because of that new users can just install nash with 'go get' and executing it just works. I know this would be something the user will need to know in the future if he wants to create their own packages, but maybe he will only use nash in the cli for awhile or just scripts and so on.

My point is: there's several new things for new users to learn, maybe we can alleviate the burden when possible.

katcipis: You are proposing to have a default NASHPATH ? Because if not I don't see a problem on make install requiring it, if the user installs with go get it will still work.

My point is that this check on make install has nothing to do with go get, I'm trying to understand the point =)

Makefile
```diff
@@ -1,8 +1,32 @@
...
+	@echo
+	@echo "Installing nash at: "$(NASHPATH)
+	mkdir -p $(NASHPATH)/bin
+	mkdir -p $(NASHPATH)/lib
+	rm -f $(NASHPATH)/bin/nash
+	rm -f $(NASHPATH)/bin/nashfmt
+	cp -p ./cmd/nash/nash $(NASHPATH)/bin
```
tiago4orion: I don't like this either. I expect that my ".nash" directory can be deleted with no harm, but in this case I'll lost my shell...
For example, when things goes wrong I commonly do something like that:

```sh
λ> cd ~
λ> mv .nash dotnash-bkp
λ> mkdir .nash
λ> nash
```
That works for every other shell, but will not work with nash

Makefile
```diff
@@ -1,8 +1,32 @@
...
+	cp -p ./cmd/nashfmt/nashfmt $(NASHPATH)/bin
+	rm -rf $(NASHPATH)/stdlib
+	cp -pr ./stdlib $(NASHPATH)/stdlib
```
tiago4orion: I know that we'd talked about this in the past but now I think we really need a second place (another env var) to put nash standard library.

hack/check.sh
```diff
@@ -33,6 +33,10 @@ done
 
 $GO tool cover -func coverage.txt
 
+echo "running stdlib tests"
+export NASHPATH=`pwd`
+./cmd/nash/nash ./stdlib/fmt_test.sh
```
tiago4orion: :+1:

stdlib/fmt_test.sh
```diff
@@ -0,0 +1,31 @@
+
+fn run_example(args...) {
+        got, status <= ./cmd/nash/nash ./stdlib/fmt_example.sh $args
+        return $got, $status
+}
+
+fn assert_success(expected, got, status) {
+        if $status != "0" {
+                exit("1")
```
tiago4orion: maybe adding a print with context about the error here, otherwise test will exit with 1 with no clue.

Makefile
tiago4orion: > You are proposing to have a default NASHPATH?

yes.

What do you think?

Makefile
tiago4orion: I know that we'd talked about this in the past but now I think we really need a second place (another env var) to put nash standard library.

Because otherwise I can mess up with my nash installation every time I messed up my .nash directory...
For example, getting a clean new nash install will be a very intricate script because I need to reset my $NASHPATH/init and other files without touching stdlib... Today this is very easy:

```sh
λ> rm -rf .nash
λ> git clone git@<someplace>/mydotnash.git .nash
λ> nash
```

tiago4orion:
People, I dislike the idea of ".nash" being our default directory for NASHPATH.
I would like to use only "nash" for this, maybe change this now is a good time.
What do you guys think?

Makefile
katcipis: Sounds good =)

Makefile
katcipis: I see, but I can't see a way to solve that and avoid having a NASHROOT

stdlib/fmt_test.sh
katcipis: To make a generic assert_success with details on the status error I will have to inject the error message on the function too because the assert function do not have enough context to give any useful detail on this case. But it is a good idea, thanks =)

Makefile
katcipis: I like the idea of a NASHROOT =)

vitorarins:
@tiago4orion Maybe NASHPATH can be `$HOME+"/nash"` and NASHROOT can be `$HOME+"/.nash"`.

Makefile
tiago4orion: Yeah, we'll need a NASHROOT but this variable could be inferred most of the time (i think).

For example, that's the install process I've in mind:

* By go get:
```sh
$ cd ~
$ go get github.com/NeowayLabs/nash/cmd/...
$ nash
λ> import "fmt"
λ> fmt_println("test")
test
λ> echo $NASHROOT
/home/user/gopath/src/github.com/NeowayLabs/nash/stdlib
```
In the scenario above, when nash is invoked it can discover that it's installed inside $GOPATH/bin and then it checks if $GOPATH/src/github.com/NeowayLabs/nash/stdlib exists and if so, then NASHROOT must be $GOPATH/src/github.com/NeowayLabs/nash/stdlib. Otherwise it must/should warn user that NASHROOT is not set and then stdlib is not available.

* By git clone && make

Golang already enforces that build must occurs inside GOPATH, then the "make install" only needs to run "go install ..." for each binary, and then everything will work as if the user had issued "go get".

* By downloading a release tarball

```sh
$ cd <someplace>
$ wget https://github.com/NeowayLabs/nash/releases/download/v0.5/nash-v0.5-linux-amd64.tar.gz
$ tar xvf nash-v0.5-linux-amd64.tar.gz -C ./nash
$ ls nash
bin
stdlib
$ ls nash/bin
nash
nashfmt
nashdoc
$ export PATH="$PATH:<someplace>/nash/bin"
$ nash
λ> import "fmt"
λ> fmt_println("test")
test
λ> echo $NASHROOT
<someplace>/nash/stdlib
```
In the case above, nash will check if it is being ran from an exploded tarball, ie., it's inside a bin directory (outside GOPATH) and directory stdlib exists in the parent folder.

* Setting NASHROOT explicitly, or install stdlib in another place

```sh
$ wget https://myserver.com/files/nash -O /usr/bin/nash
$ chmod +x /usr/bin/nash
$ nash
λ> import "fmt"
ERROR: <interactive>:1:0: Failed to import path 'fmt'. The locations below have been tried:
 "./fmt", "./fmt.sh", "/home/i4k/.nash/lib/fmt", "/home/i4k/.nash/lib/fmt.sh"
λ> 
λ> curl -s https://myserver.com/files/stdlib.tar.gz | tar xvf -C $HOME+"/stdlib" -
λ> setenv NASHROOT = $HOME+"/stdlib"
λ> nash
λ> import "fmt"
λ> fmt_println("test")
test
λ> 
```

If someone just don't want the stdlib, or want to put it somewhere else, or he just wants to drop the shell inside the /usr/bin of a server, then it could just use the shell without stdlib or he can set explicitly the NASHROOT to some place where it exists.

tiago4orion:
@vitorarins I dislike the idea of "hidden directories" at all. I made a proposal here:
https://github.com/NeowayLabs/nash/pull/228#discussion_r126682277
What do you think?

Makefile
katcipis: Im not very fond to this idea:

```
In the scenario above, when nash is invoked it can discover that it's installed inside $GOPATH/bin and then it checks if $GOPATH/src/github.com/NeowayLabs/nash/stdlib
```

These days Go can even work with default GOPATH (there will be no env var exported), it seems like we are mixing nash with Go too much =/. I was more prone to make it harder to install with go get but keep it more simple, no logic will involve GOPATH.

If you go get nash, it will work with default NASHPATH and NASHROOT, but there will be no stdlib available. If you want the stdlib you must:

* Install with make
* Install release with tarball
* Do it manually copying stdlib to NASHROOT

I never liked Go thing of only building inside GOPATH... I like being able to build and run nash tests where I want =), specially because I don't see a huge benefit on enforcing this.

tiago4orion: > These days Go can even work with default GOPATH (there will be no env var exported)

Yes, and then we can check if it's inside GOPATH or inside $HOME/go :stuck_out_tongue: 

> it seems like we are mixing nash with Go too much =/. I was more prone to make it harder to install with go get but keep it more simple, no logic involving GOPATH.

When you install with "go get" you aren't just installing but compiling from source code. Nash is written in Go, then is very appropriate to easy the installation process for people using nash from source code.

'Go get' is the first thing that came to mind when someone sees a Go project, then it's reasonable to make it work in the first place. I enjoy go-gettable projects.. would be sad that nash didn't work that way.

> I already disliked how my print implementation leaked Go's details, this seems to enforce it even more (what does my nash have to do with GOPATH ?). Ideally the user could never even know that nash is written in Go.

About the `print` implementation leaking Go's internal format, it was a implementation design choice. In the future, that do not need to be that way.
Users that download from binary will never know that nash is written in Go... Unless they look inside the binary to find GOPATH string there or trace the system calls to know that we probe golang directories.. But if someone hexdump the binary he will see the runtime symbols anyway (if not striped).

I didn't say that we need to add docs to users rely on GOPATH... or put nash inside GOPATH and things like that. I'm only saying that we can make the life easier for people installing from source code. I love when things just works.

But I can live with that. Your call, we can experiment that way and discuss in the future if this turns out a problem.

katcipis: You said this:

```
In the scenario above, when nash is invoked it can discover that it's installed inside $GOPATH/bin and then it checks if $GOPATH/src/github.com/NeowayLabs/nash/stdlib exists and if so, then NASHROOT must be $GOPATH/src/github.com/NeowayLabs/nash/stdlib. Otherwise it must/should warn user that NASHROOT is not set and then stdlib is not available.
```

If this is the behavior we will have to document that, so how we will not "add docs to users rely on GOPATH" ?

I think the main problem is that I don't see nash as much as a Go project, it is big enough for me to see as something decoupled from Go, even on how I get the binaries. If I want to build from master, we have make install, which is a more universal way to build stuff. I enjoy go get more for small tools. But to be honest this is a feeling that may not even make sense...I don't know, Kubernetes is also pretty decoupled from the fact that it is developed in Go when you deploy it, but I'm not sure if it is based only on the size of the project.

I just feel clumsy to document these things to the user, and if the behavior will be that we must to it (I hate implicit undocumented behavior).

tiago4orion: We could document something like:

--- snip start ---
The environment variable NASHROOT points to the location of the stdlib in the system. If it is not set explicitly by user, then nash will try to find the stdlib in common places. If you're curious about where's the location of the stdlib in use by current shell session, simply type:

```sh
λ> echo $NASHROOT
/home/i4k/go/src/github.com/NeowayLabs/nash/stdlib
```
--- snip end ---

tiago4orion: but I'm ok with your idea. We can keep this way.

katcipis: What about this:

1 - If NASHROOT is set, use it
2 - If not, use $HOME/nash
3 - If $HOME/nash do not have stdlib, use $GOPATH/github.com/NeowayLabs/nash
4 - Explode in million pieces if everything fails

I like using the make install because if I delete my entire GOPATH my nash still works perfectly, it is totally decoupled from my Go installation. But this way it still works to people who like to couple it with Go =)

tiago4orion: SGTM

katcipis:
I was doing some code today, first of all I set the NASHROOT default to $HOME/nashroot, so it is safe to remove the default NASHPATH too (like @tiago4orion wanted, make it easy to delete all 3rdparty libs but dont fuck up the nash install).

But I got thinking on what to do when everything fails. Like when there is no NASHPATH and no HOME env var (same thing to NASHROOT + HOME + GOPATH). Should we just try to import the file and crash or give an error like "No env vars setted, go set them" ? =P (I'm prone to just trying to import, but this may make debugging when there is no env var available harder).

tiago4orion:
@katcipis I'm not sure either, but:

```
λ> import something
ERROR: <interactive>:1:0: Failed to import path 'something'. The locations below have been tried:
 "./something", "./something.sh", "/home/i4k/.nash/lib/something", "/home/i4k/.nash/lib/something.sh"
```

Maybe only trying to import is good, but improve the error message saying that NASHROOT is not set?

katcipis:
Trying to import and improved error message seems the way to me, like @tiago4orion suggested. Anyone opposes ?

Makefile
```diff
@@ -1,8 +1,32 @@
...
+
+guard-%:
+	@ if [ "${${*}}" = "" ]; then \
+                echo "'$*' must be set in order to install and use nash"; \
+                exit 1; \
+        fi
+
+install: build guard-NASHPATH
+	@echo
+	@echo "Installing nash at: "$(NASHPATH)
```
tiago4orion: this isn't the behaviour we had discussed. Or am I wrong?

README.md
```diff
@@ -50,14 +56,34 @@ Nash is a system shell, inspired by plan9 `rc`, that makes it easy to create rel
 
 ## Installation
 
-If you have Go, go-get it:
+Before proceeding to any of the install procedures it is important that you
```
tiago4orion: Where will the user set this environment variable? I think this is confusing because nash will init their environment from `$NASHPATH/init`, but NASHPATH needs to be set in the first place... But how if nash is the login shell? You mean to place the NASHPATH env config inside /etc/profile ?

README.md
```diff
@@ -50,14 +56,34 @@ Nash is a system shell, inspired by plan9 `rc`, that makes it easy to create rel
...
-```sh
-# Make sure GOPATH/bin is in your PATH
-go get github.com/NeowayLabs/nash/cmd/nash
-go get github.com/NeowayLabs/nash/cmd/nashfmt
+Also add the directory **$NASHPATH/bin** to your **PATH** so you can run
+nash in a more practical way (although not necessary).
+
+### Release
+
+Installing is so stupid that we provide small scripts to install it.
```
tiago4orion: I do not understand this section. The user will download the source code to then invoke a script to download and install a release (binary) one?

hack/install/linux.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
tiago4orion: what's the use case for this script?

internal/sh/shell.go
```diff
@@ -752,6 +752,65 @@ func (shell *Shell) executeReturn(n *ast.ReturnNode) ([]sh.Obj, error) {
 	return returns, newErrStopWalking()
 }
 
+func (shell *Shell) getNashRootFromGOPATH(preverr error) (string, error) {
```
tiago4orion: NASHROOT is the name of the env var, but internally we can call this path only `root`. Then the function names could be something like `sh.rootFromGOPATH()`, `sh.isValidRoot(path)` and create a public one called `sh.NashRoot()`. What do you think?
I'm not sure if this is good, feel free to ignore.

internal/sh/shell.go
```diff
@@ -752,6 +752,65 @@ func (shell *Shell) executeReturn(n *ast.ReturnNode) ([]sh.Obj, error) {
...
+	return gopath + "/src/github.com/NeowayLabs/nash", nil
+}
+
+func isValidNashRoot(nashroot string) bool {
+	_, err := os.Stat(nashroot + "/stdlib")
+	return err == nil
+}
+
+func (shell *Shell) getNashRoot() (string, error) {
+	// TODO(katcipis): It is very annoying to load env vars, perhaps a shell.GetStringEnv ?
```
tiago4orion: makes sense, maybe we can export a `Getenv` that returns a empty string if not found, but use an unexported one internally, like `getenv` that returns a bool indicating if the env does exists ? 

We need the bool one internally because nash have to make sure the env exists and isn't only set to "".

You can just discards the bool in the call of `Getenv`

internal/sh/shell.go
```diff
@@ -798,20 +857,17 @@ func (shell *Shell) executeImport(node *ast.ImportNode) error {
 		}
 	}
 
-	nashPath, ok := shell.Getenv("NASHPATH")
-
-	if !ok {
-		return errors.NewError("NASHPATH environment variable not set on shell %s", shell.name)
-	} else if nashPath.Type() != sh.StringType {
-		return errors.NewError("NASHPATH must be n string")
+	dotDir, nashpatherr := shell.getNashPath()
```
tiago4orion: libs installed in NASHPATH have precedence over NASHROOT?

tiago4orion: just asking, I'm not sure what's the right behaviour

internal/sh/shell_env_test.go
```diff
@@ -0,0 +1,269 @@
...
+	"os"
+	"testing"
+)
+
+func TestLoadsLibFromNASHPATH(t *testing.T) {
+	home, teardown := setupEnvTests(t)
+	defer teardown()
+
+	nashpath := home + "/testnashpath"
+	os.Setenv("NASHPATH", nashpath)
```
tiago4orion: As I'd commented in the docs file, I found good this case works this way but for real use it'll require another login shell being used to set this env var or maybe put this environment in the /etc/profile.d of the machine. 

I'm not complaining about the test, only want to know from you what's your expected use case for that.

README.md
tiago4orion: I thought more about this and I think is useful for this behaviour to work when someone wants to put their NASHPATH somewhere else other than the default path. But I think it should be the default way of installing nash, because it requires another login shell or root access to /etc/profile.d

internal/sh/shell_env_test.go
tiago4orion: But now talking about the test =P
I don't think is good to set env vars in test. First because even though it's safe to call os.Setenv concurrently, it is not between two tests depending on changes in the environment (it will require sync between tests). Another problem is that even if we add proper locks in tests, this will require to restore the environment in a teardown in each test.

The Go environ is *not* the process environ, but just a map[string]string maintained in the runtime and protected by an envLock. I found this very surprisingly nice, but the environ belong to the shell and nash also has a map of their environments and we do not change the runtime environ every time our environ is changed... Well, to avoid problems, better to use the environ for their purpose, passing it around to child processes.

Maybe I've used this pattern in other tests, I'm not sure, but only remember that this is not good now,

A better approach should be to make `newTestShell` executing the test snippet in a new process and pass the environment variables explicitly to the child process (by os/exec.Cmd.Environ).

tiago4orion: I'm not saying we have to make every test run in a child process, but only the ones depending on changes in the environment.

internal/sh/shell_env_test.go
```diff
@@ -0,0 +1,269 @@
...
+	err = os.Setenv("HOME", home)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	mkdirAll(t, home)
+
+	return home, func() {
+		errs := []error{}
+		errs = append(errs, os.Setenv("HOME", curhome))
```
tiago4orion: this mess of teardown could be avoided by passing the environment variables explicitly to the child process.

internal/sh/shell_test.go
```diff
@@ -23,29 +23,45 @@ type execTestCase struct {
...
+	gopath := os.Getenv("GOPATH")
 
 	if gopath == "" {
-		panic("Please, run tests from inside GOPATH")
+		t.Fatal("Please, run tests from inside GOPATH")
 	}
 
-	testDir = gopath + "/src/github.com/NeowayLabs/nash/" + "testfiles"
-	nashdPath = gopath + "/src/github.com/NeowayLabs/nash/cmd/nash/nash"
+	testDir := gopath + "/src/github.com/NeowayLabs/nash/" + "testfiles"
```
tiago4orion: we are repeating "github.com/NeowayLabs/nash" in several places. I don't think this must be made now, but maybe we can add a constant string pointing to this and reuse this "name". In the future we can change to a [canonical import path](https://golang.org/doc/go1.4#canonicalimports) like "neoway.io/nash" and the change could be less troubled.

internal/sh/shell_test.go
```diff
@@ -23,29 +23,45 @@ type execTestCase struct {
...
-	testDir, gopath, nashdPath string
-)
+type fixture struct {
+	dir       string
+	nashdPath string
+}
 
-func init() {
-	gopath = os.Getenv("GOPATH")
+func setup(t *testing.T) (fixture, func()) {
```
tiago4orion: much better :+1:

Makefile
katcipis: missing a fix for this

README.md
katcipis: If you wish to improve this docs it will be welcomed, I think you understand this better =)

hack/install/linux.sh
katcipis: Installing a nash release (instead of the master)

internal/sh/shell.go
katcipis: Why a public one ?

internal/sh/shell.go
katcipis: The problem is not existence checking, it is the type validation to see if the env var is a string or not, that is annoying =P. In the end I dont check and let it panic if the variable is not a string.

internal/sh/shell.go
katcipis: NASHROOT is used only to search for stdlib, there will be no lookup at NASHROOT/lib

internal/sh/shell_env_test.go
katcipis: I dont use nash as the default shell, I always login on zsh and later I call nash =)

katcipis: I followed the idea that was here:

https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell_test.go#L44

But I wanted to isolate each test, so I set the env vars and use defer to restore them. Executing it on different processes seems better indeed.

internal/sh/shell_env_test.go
katcipis: dont find it that messy, but dont like manipulating global state

katcipis: this is messy =)

https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell.go#L782

hack/install/linux.sh
tiago4orion: but someone will download the entire source code to just invoke a script that downloads the binary release?
Or you mean to installations like below?

```sh
$ curl https://raw.githubusercontent.com/NeowayLabs/nash/master/hack/install/linux.sh | bash -
```
If this is the case, then better to update the README explaining the right way of invoking this script.

internal/sh/shell.go
tiago4orion: consistency: http://godoc.org/github.com/NeowayLabs/nash#Shell.DotDir

and this could be useful when embedding nash in Go.

but feel free to ignore, we can add in the future if needed.

internal/sh/shell.go
tiago4orion: ah, got it. That's because nash environment vars could be other types other than strings.... we need to review this, because I've not see any advantage.

internal/sh/shell.go
tiago4orion: I mean:

```
import "os/file"
```

must look first in $NASHPATH/lib or $NASHROOT/stdlib ?
The current code seems to look first in NASHPATH/lib.

internal/sh/shell_env_test.go
tiago4orion: got it, my use case is tested in function `TestLoadsLibFromHOMEIfNASHPATHIsUnset` then kkk :joy: 
But is good to know this works anyway if I want to change default location in the future.

tiago4orion: Setenv on init is safe, but I'm pretty sure I've made something similar on other tests. 
Maybe just leave this change to the future, I can address this in my next work on nash.

internal/sh/shell_env_test.go
tiago4orion: really? You're just manipulating global state here... :joy: 
I wasn't referring to the teardown function, but to the global state being manipulated in a unsafe way.

Regarding the link, sure it can be improved. It was a premature optimization, sorry.

internal/sh/shell.go
katcipis: Good question, no thoughts on that yet

katcipis:
@tiago4orion is it good enough to merge ? this way we can evolve it on smaller steps (the PR is pretty big already)

tiago4orion:
@katcipis yes, but can I change the "make install" to require NASHROOT to be set? and install stdlib there?

I'm thinking in something like:

```
install: build guard-NASHROOT
    @echo
    @echo "Installing nash at: $GOPATH/bin"
    @echo "Installing stdlib at: $NASHROOT"
    go install ./cmd/...
    cp -pr ./stdlib $(NASHROOT)/stdlib
```

Only for now, we can improve this in the next PRs.
What do you think?

katcipis:
@tiago4orion I fixed that already and I could swear I pushed it...but now I see that I didnt, let me find the changes and push them =P

katcipis:
@tiago4orion done. I also added default values like you suggested, so it will work without any env var set (except HOME)

PR #231, opened 2017-08-05
Add proposal for scope management proposal

katcipis:


proposal/1-scope-management.md
```diff
@@ -0,0 +1,71 @@
...
+remains empty, why is that ? The problem is on the add function:
+
+```sh
+fn add(val) {
+        l <= append($l, $val)
+}
+```
+
+When we reference the **l** variable it uses the reference on the
+outer scope (the empty list), but there is no way to express syntactically
```
ppizarro: Wouldn't be: local scope?

proposal/1-scope-management.md
```diff
@@ -0,0 +1,71 @@
...
+that we want to change the list on the outer scope instead of creating
+a new variable **l**. That is why the **get** and **print** functions
+are always referencing an outer list **l** that is empty, a new one
```
ppizarro: Wouldn't be: "are always referencing an new local list **l** that is empty..."

proposal/1-scope-management.md
katcipis: Actually I was trying to say that when you reference with $l it will access the outer scope **l**, but you can't change it, when you try you end up creating a new **l**

proposal/1-scope-management.md
katcipis: Actually no, when you call **get** and **print** they work because they reference the outer scope **l**. A local scope **l** is only created on the **add** function. The local scope **l** created at the **add** function is never empty since it receives the result of appending the outer scope **l** with he value being added.

ppizarro: uauauauaua... This is a bit confusing.
If I understand correctly when I read the variable I use the scope of the parent (outer) but when I do an assignment it creates a new one in the local scope and doesn't use the one of the parent. Is it?

katcipis: @ppizarro exactly

What is confusing is that on this line:

```
l <= append($l, $val)
```
The following happens:

* You read the outer scope **l** (empty)
* append $val to it (list with $val inside)
* create a new list **l** that will be ($val)

There is no way today to:

* read the outer scope **l** (empty)
* append $val to it (list with $val inside)
* update the outer scope **l** to be ($val)

Indeed not very simple, I'm struggling to give good examples and to understand it in details myself =P

tiago4orion:
Hello guys,

Just added some text elaborating the idea of each proposal. Feel free to change if there's some mistake.

What are the cons of the `outer` proposal?

proposal/1-scope-management.md
```diff
@@ -0,0 +1,224 @@
...
+confuse because it declares new variables **and** update existent ones
+(in outer scopes also). Then making hard to know what variables are
+being declared local and what are being updated, by just looking at
+the statement, because the meaning will depend in the current
+environment of variables.
+
+Another downside of `var` is their very incompatible nature. Every
+nash script ever created will be affected.
+
+## Proposal II - "outer"
```
vitorarins: I know this is really personal, but I don't feel good about this name. 

I would like to propose changing this to `ext` for external. What do you think?

tiago4orion: I have nothing against "outer", but found weird that few languages chose this name.
`Ext` is too cryptic for my taste, it lacks the meaning and could confuse with `extension`.

tiago4orion: Using google I found that `outer`is used in languages *D*, *SQL* and *gosu*.

katcipis: I prefer outer than ext (or global, like python)

ppizarro: I also didn't like it very much but I think that of all the names proposed was the best.

Another thing I don't like about this proposal is that if in the future we have block scope the code will get full of "outer".

ppizarro:
cons: 'outer' proposal (in my opinion)

- It is confusing as to the scope of the variable:
```sh
x = "1"

func test() {
    // this line is confusing: read from outer scope but initialize a new variable in local scope
    x <= inc($x)
}
```

- Another thing I don't like about this proposal is that if in the future we have block scope the code will get full of "outer".

proposal/1-scope-management.md
```diff
@@ -0,0 +1,239 @@
...
+confuse because it declares new variables **and** update existent ones
+(in outer scopes also). Then making hard to know what variables are
+being declared local and what are being updated, by just looking at
+the statement, because the meaning will depend in the current
+environment of variables.
+
+Another downside of `var` is their very incompatible nature. Every
+nash script ever created will be affected.
+
+Another behavior we need to discuss is whether all variables declared
```
tiago4orion: I don't get if this refers to implementation or var's semantics. For the same code example, you mean that in an alternative design it could print an empty string?

proposal/1-scope-management.md
vitorarins: >Ext is too cryptic for my taste, it lacks the meaning and could confuse with extension

@tiago4orion I see the confusion now.

 I've seen that `C` has a keyword `extern`, maybe that would be better, I don't have a strong opinion here, just sharing some ideas.

But `outer` is not bothering me that much anymore. You can ignore this comment.

proposal/1-scope-management.md
ppizarro: I think this refers to var's semantics.
Yes, in an alternative design it could print an empty string (like javascript) but I don't like this. We could argue about it if someone has a contrary opinion.

tiago4orion: :+1:

PR #234, opened 2017-08-10
Small fixes on proposal

katcipis:


vitorarins:
@katcipis I thought you were going to refactor the part about downsides of the `var` statement. Specifically the part saying about the implementation we have today in the PR. Teorically we are going to change the behaviour a little bit when all of the variables already exist.

katcipis:
@vitorarins actually I don't see any upside on having the behavior where re-declaring fails... but if you guys think it is worth it I can do it.

PR #235, opened 2017-08-18
Add support to init.sh

katcipis:
I enjoy always using .sh but it did not worked on init, it seemed consistent to support this =)

### DIFF
--- a/cmd/nash/cli.go
+++ b/cmd/nash/cli.go
@@ -51,16 +51,36 @@ func execFn(shell *nash.Shell, fnDef sh.FnDef, args []sh.Obj) {
 	}
 }
 
-func setupCli(shell *nash.Shell) error {
-	shell.SetInteractive(true)
-	initFile := shell.DotDir() + "/init"
+func importInitFile(shell *nash.Shell, initFile string) error {
 
-	if d, err := os.Stat(initFile); err == nil && !noInit {
+	if d, err := os.Stat(initFile); err == nil {
 		if m := d.Mode(); !m.IsDir() {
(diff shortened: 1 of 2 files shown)


cmd/nash/cli.go
```diff
@@ -67,6 +65,28 @@ func setupCli(shell *nash.Shell) error {
...
+func setupCli(shell *nash.Shell) error {
+	shell.SetInteractive(true)
+
+	if noInit {
+		return nil
+	}
+
+	initFiles := []string{
+		shell.DotDir() + "/init",
+		shell.DotDir() + "/init.sh",
```
tiago4orion: Both files will be imported if both exists?

tiago4orion:
I'd no idea that the init file doesn't work with ".sh" extension... 
I think that "init" must be handled in the same way as other imports, no?

Don't you think the right solution should be fixing the `import` eval to handle that correctly?
https://github.com/NeowayLabs/nash/blob/master/internal/sh/shell.go#L848

cmd/nash/cli.go
katcipis: @tiago4orion yay you are alive =), missed u.

Putting feelings apart, yeah they both will, I did not liked it but it was the fastest thing I could do to make it work.

In my opinion in face of an ambiguity it is better to fail, just importing one of them by some arbitrary order seems wrong, but I was in a rush to get things working and wanted to see if it would work =P

katcipis:
@tiago4orion I think you are right on your remarks, Im just thinking if we cant do the improvements you suggested on another MR, right now init.sh does not even work and Im not sure when I will be able to make the improvements.

tiago4orion:
Sure!

PR #237, opened 2017-08-21
Some tweaks to work on windows

### LINKED ISSUE #236: windows: Use proper env variable to set NASHPATH
image

To start nash on windows requires setting `NASHPATH` explicitly, but we can use the environment variable `USERPROFILE` or `HOMEPATH`.

tiago4orion:
This PR makes the following changes:

- Use portable API to get user's home directory. Closes #236 
- Use portable API for path manipulation;
- Fix format error message containing "%Path%"

Nash now is pretty usable on windows.


cmd/nash/cli.go
```diff
@@ -52,21 +51,18 @@ func execFn(shell *nash.Shell, fnDef sh.FnDef, args []sh.Obj) {
 	}
 }
 
-func setupCli(shell *nash.Shell) error {
-	shell.SetInteractive(true)
-	initFile := filepath.Join(shell.DotDir(), "init")
-
+func importInitFile(shell *nash.Shell, initFile string) (imported bool, err error) {
 	if d, err := os.Stat(initFile); err == nil {
 		if m := d.Mode(); !m.IsDir() {
```
katcipis: Could be d.Mode().IsDir() (not sure if it is better, pending to thinking it is a little easier on the eyes).

cmd/nash/main.go
```diff
@@ -102,18 +103,21 @@ func getnashpath() (string, error) {
...
-		} else {
-			return "", fmt.Errorf("Environment variable NASHPATH or $USER must be set")
-		}
+	usr, err := user.Current()
+	if err != nil {
+		return "", fmt.Errorf("unable to infer nash directory. "+
+			"You must set NASHPATH environment variable explicity: ",
+			err.Error())
+	}
+	if usr.HomeDir == "" {
```
katcipis: Why compare usr.HomeDir with "" twice ?

cmd/nash/main.go
```diff
@@ -102,18 +103,21 @@ func getnashpath() (string, error) {
...
+		return "", fmt.Errorf("User %s doesn't have a home directory. "+
+			"Set NASHPATH environment variable to desired location.", usr.Name)
+	}
+	if usr.HomeDir == "" {
```
katcipis: Here is the second compare

errors/error.go
```diff
@@ -43,7 +43,11 @@ func NewEvalError(path string, node ast.Node, format string, arg ...interface{})
 }
 
 func (e *NashError) SetReason(format string, arg ...interface{}) {
-	e.reason = fmt.Sprintf(format, arg...)
+	if len(arg) > 0 {
+		e.reason = fmt.Sprintf(format, arg...)
```
katcipis: Just calling Sprintf wont work ? that is sad =/

internal/sh/cmd.go
```diff
@@ -45,7 +46,7 @@ func NewCmd(name string) (*Cmd, error) {
 
 	cmd := Cmd{}
 
-	if name[0] != '/' {
+	if !filepath.IsAbs(name) {
```
katcipis: Nice =)

internal/sh/shell.go
```diff
@@ -1209,14 +1197,8 @@ func (shell *Shell) buildRedirect(cmd sh.Runner, redirDecl *ast.RedirectNode) ([
 			cmd.SetStdout(file)
 			closeAfterWait = append(closeAfterWait, file)
 		case ast.RedirMapSupress:
-			file, err := os.OpenFile("/dev/null", os.O_RDWR, 0644)
-
-			if err != nil {
-				return closeAfterWait, err
-			}
-
+			file := ioutil.Discard
```
katcipis: nice

katcipis:
Ignore when I said "Why the name is "fix-windows" ? Or am I missing something ?", github interface trolled me =)

cmd/nash/main.go
tiago4orion: Probably because I'm dumb.

cmd/nash/cli.go
tiago4orion: 👍

errors/error.go
tiago4orion: I had the same doubt, but implemented this way because had no time to test/verify. I'll take a look.

tiago4orion: yeah, it works =)

internal/sh/shell.go
```diff
@@ -412,14 +412,9 @@ func (shell *Shell) setupBuiltin() {
...
 	err := shell.Exec(shell.name, `fn nash_builtin_cd(path) {
-            if $path == "" {
-                    path = $HOME
-            }
-
-            chdir($path)
-        }
-
-        bindfn nash_builtin_cd cd`)
+		chdir($path)
```
tiago4orion: I made this in a rushed to make things work... But this is very very bad... It actually broke the tests also..

On Windows, the $HOME variable don't exists. What we can do here? Go code now is portable, but nash doesn't have portable environment variables (yet?).

tiago4orion: @katcipis @vitorarins What do you guys think?

katcipis: @tiago4orion I'm not understanding exactly what do you mean by "nash doesn't have portable environment variables".

For the problem of "Windows not having HOME" we could add some code to load different env vars according to the OS, like when it is on windows it loads from "WHATEVERWINDOWSCRAP", on posix "HOME", don't see a way around doing something like this =/.

tiago4orion: by portable environment variables I meant exposing the environment vriables HOME, USER, and so on, to nash scripts regardless of the operating system running (backed by different environ in each OS). Then, my `cd` bindfn function could use PATH environ and be portable. I think this is what you have proposed, but I'm not sure.

But I dont know if this is a good solution, because only some variables will be portable... and still some os-dependent code will be required in some cases.

Maybe the best solution is using a different default `cd` bindfn for windows guaranteeing the same semantics but using a different env var:

```
fn __nash_buildtin_cd(path) {
    if $path == "" {
        chdir($WHATEVERWINDOWSCRAP)
    }
    # etc
}
```

katcipis: Now I understand, I thought it was something with env vars that was not working properly, you are proposing to get a Windows env var like $SOMESHIT and map it to $HOME under the hood transparently to the user ?

tiago4orion: Yes, but now I dont think this is good anymore kkk =(

katcipis: I thought a little about it and it does not seem like a good idea, we would need to set a boundary to how far we go with this, would it be just HOME ? posix stuff ? XDG ? I'm not sure, but I'm more prone to less, if we are wrong it is easier to add it later.

katcipis: Wow, I just commented that before seeing your comment =P

vitorarins: What if we: on Windows, add an env var `HOME` that points to `C:\Users\$USER` (I am not sure what is the default on Windows, but I think it looks something like this)

tiago4orion: @katcipis Yeah, same feelings.

@vitorarins and where $USER come from? I'm not saying this is bad, but we need to decide if we will deliver a POSIX shell interface in every operating system supported or we will add support for conditional eval of scripts (like Go build flags, or C ifdef and so on).

I really don't know where we will end in both cases...

katcipis: I find conditional eval per operational system more useful in general, trying to deliver full POSIX on every OS seems more cumbersome and prone to not solve all the problems anyway

tiago4orion: I prefer conditonal eval also, but not sure in how it can be done. Ideas ?

tiago4orion:
:x: [Build nash 1.0.0.29 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.29) (commit https://github.com/NeowayLabs/nash/commit/3b98f8c4a6 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.31 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.31) (commit https://github.com/NeowayLabs/nash/commit/3f6400e908 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.33 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.33) (commit https://github.com/NeowayLabs/nash/commit/d12d2e1c3a by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.42 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.42) (commit https://github.com/NeowayLabs/nash/commit/6567e6df1f by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.44 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.44) (commit https://github.com/NeowayLabs/nash/commit/c0cf15e42f by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.46 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.46) (commit https://github.com/NeowayLabs/nash/commit/aeb567c1f2 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.48 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.48) (commit https://github.com/NeowayLabs/nash/commit/7ae3a157ba by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.51 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.51) (commit https://github.com/NeowayLabs/nash/commit/14fa026299 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.53 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.53) (commit https://github.com/NeowayLabs/nash/commit/b625a8c553 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.55 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.55) (commit https://github.com/NeowayLabs/nash/commit/33675b8682 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.57 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.57) (commit https://github.com/NeowayLabs/nash/commit/8a3846f074 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.59 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.59) (commit https://github.com/NeowayLabs/nash/commit/2efa1c03d1 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.61 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.61) (commit https://github.com/NeowayLabs/nash/commit/45974f87e9 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.63 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.63) (commit https://github.com/NeowayLabs/nash/commit/0d0a86036b by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.65 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.65) (commit https://github.com/NeowayLabs/nash/commit/f6c66f33dd by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.67 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.67) (commit https://github.com/NeowayLabs/nash/commit/45ca90b0ec by @tiago4orion)

tiago4orion:
@katcipis @vitorarins There's only one test failing now and it's a Go <= 1.9 issue:
https://github.com/golang/go/issues/22278
This is already fixed in Go's master branch.

I'm going to merge this now. Anyone against this, please, speak now or forever hold your peace :-)

appveyor.yml
```diff
@@ -0,0 +1,24 @@
+version: "1.0.0.{build}"
```
katcipis: What is this file ? o.O

ast/node.go
```diff
@@ -289,15 +289,6 @@ type (
 		cmdname string
 	}
 
-	// A DumpNode represents the "dump" keyword.
```
katcipis: dump has been removed ? this MR wasn't just for windows stuff ? o.O

cmd/nash/cli.go
```diff
@@ -51,18 +51,18 @@ func execFn(shell *nash.Shell, fnDef sh.FnDef, args []sh.Obj) {
...
-			err = shell.ExecuteString("init", `import "`+initFile+`"`)
+			err = shell.ExecuteString("init",
+				fmt.Sprintf("import %q", initFile))
 			if err != nil {
-				return fmt.Errorf("Failed to evaluate '%s': %s\n", initFile, err.Error())
+				return false, fmt.Errorf("Failed to evaluate '%s': %s\n", initFile, err.Error())
```
katcipis: I find confusing to mix named returns with normal returns, not terrible, perhaps a matter of taste

cmd/nash/cli.go
```diff
@@ -78,10 +78,13 @@ func setupCli(shell *nash.Shell) error {
 	}
 
 	for _, init := range initFiles {
-		err := importInitFile(shell, init)
+		imported, err := importInitFile(shell, init)
 		if err != nil {
 			return err
 		}
+		if imported {
+			break
```
katcipis: It may be the odd github MR interface, but it seems that here it could be just ```return nil``` ?

hack/check.sh
```diff
@@ -16,7 +16,7 @@ GO="go"
 echo "mode: count" > coverage.txt
 
 # Standard $GO tooling behavior is to ignore dirs with leading underscors
-for dir in $(find . -maxdepth 10 -not -path './.git*' -not -path './Godeps*' -not -path './contrib*' -not -path './cmd/nash/vendor*' -not -path './research*' -type d);
+for dir in $(find . -maxdepth 10 -not -path './.git*' -not -path './cmd/nash/vendor*' -type d);
```
katcipis: If we are going for Go > 1.9 I think vendor directories area already ignored. Not sure if it is just that that is being done here althought

internal/sh/shell.go
```diff
@@ -411,21 +401,23 @@ func (shell *Shell) setupBuiltin() {
 
 func (shell *Shell) setupDefaultBindings() error {
 	// only one builtin fn... no need for advanced machinery yet
-	// FIXME: seems better to have a std init than embedding here on code
-	err := shell.Exec(shell.name, `fn nash_builtin_cd(args...) {
+	homeEnvVar := "HOME"
```
katcipis: Why not use the "user" package here to get the home dir ?

internal/sh/shell.go
```diff
@@ -760,26 +744,25 @@ func (shell *Shell) getNashRootFromGOPATH(preverr error) (string, error) {
...
 	return err == nil
 }
 
 func (shell *Shell) getNashRoot() (string, error) {
 	// TODO(katcipis): It is very annoying to load env vars, perhaps a shell.GetStringEnv ?
 
 	nashroot, ok := shell.Getenv("NASHROOT")
-
 	if !ok {
 		h, hashome := shell.Getenv("HOME")
```
katcipis: Here we also need the user package no ? Perhaps we could use a helper function to load this using user

internal/sh/shell.go
```diff
@@ -852,24 +828,25 @@ func (shell *Shell) executeImport(node *ast.ImportNode) error {
...
 		if !hasExt {
-			tries = append(tries, path.Dir(shell.filename)+"/"+fname+".sh")
+			tries = append(tries, localFile+".sh")
 		}
 	}
 
 	dotDir, nashpatherr := shell.getNashPath()
 	if nashpatherr == nil {
-		tries = append(tries, dotDir+"/lib/"+fname)
+		tries = append(tries, filepath.Join(dotDir, "lib", fname))
```
katcipis: Instead of always having to check if it has an extension before adding on "tries" wouldn't be better to the code that used "tries" to handle this only in one place by checking if the extension exists there ?

internal/sh/shell_regression_test.go
```diff
@@ -5,34 +5,38 @@ import (
...
 func TestExecuteIssue68(t *testing.T) {
 	sh, err := NewShell()
-
 	if err != nil {
 		t.Error(err)
 		return
 	}
 
-	err = sh.Exec("-input-", `echo lalalala | grep la > /tmp/la`)
+	tmpDir, err := ioutil.TempDir("", "nash-tests")
```
katcipis: nice =D

nash_test.go
```diff
@@ -19,14 +24,25 @@ func init() {
 	gopath = os.Getenv("GOPATH")
 
 	if gopath == "" {
-		panic("Please, run tests from inside GOPATH")
+		usr, err := user.Current()
```
katcipis: This snippet is the code that we could extract to use both on tests and nash code

spec_test.go
```diff
@@ -5,14 +5,27 @@ import (
...
+	"path/filepath"
 
 	"golang.org/x/exp/ebnf"
 )
 
 func TestSpecificationIsSane(t *testing.T) {
-	filename := os.Getenv("GOPATH") + "/src/github.com/NeowayLabs/nash/spec.ebnf"
+	gopath := os.Getenv("GOPATH")
+	if gopath == "" {
+		usr, err := user.Current()
```
katcipis: YAHC (Yet Another Home Code) =P

appveyor.yml
tiago4orion: appveyor.com

ast/node.go
tiago4orion: I merged the other branch here to avoid having to fix 'dump' keyword issues on windows.

cmd/nash/cli.go
tiago4orion: I don't why this function has named returns haha

cmd/nash/cli.go
tiago4orion: one success exit is not better?

cmd/nash/main.go
tiago4orion: forgot this. Thanks!

tiago4orion:
:x: [Build nash 1.0.0.73 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.73) (commit https://github.com/NeowayLabs/nash/commit/a94348bc5e by @tiago4orion)

hack/check.sh
tiago4orion: we test each file separately because of the coverage profile, then we need to skip vendor files here.

tiago4orion:
:x: [Build nash 1.0.0.75 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.75) (commit https://github.com/NeowayLabs/nash/commit/73f06219f3 by @tiago4orion)

internal/sh/shell.go
tiago4orion: good!

tiago4orion:
:x: [Build nash 1.0.0.77 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.77) (commit https://github.com/NeowayLabs/nash/commit/291bab9d01 by @tiago4orion)

tiago4orion:
:x: [Build nash 1.0.0.79 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.79) (commit https://github.com/NeowayLabs/nash/commit/454c3c86cf by @tiago4orion)

cmd/nash/cli.go
vitorarins: @tiago4orion, the `noInit` option is already `return nil` and it is technically a "success" exit too, right?

internal/sh/shell.go
tiago4orion: probably yes, but the import logic has few tests and some know problems already. Better to land this change in another PR with proper tests and discussion about the import order semantics.
this code definitely isn't good.

internal/sh/shell.go
```diff
@@ -411,23 +402,25 @@ func (shell *Shell) setupBuiltin() {
 
 func (shell *Shell) setupDefaultBindings() error {
 	// only one builtin fn... no need for advanced machinery yet
-	// FIXME: seems better to have a std init than embedding here on code
-	err := shell.Exec(shell.name, `fn nash_builtin_cd(args...) {
+
+	usr, err := user.Current()
```
tiago4orion: well, now I'm thinking if this makes sense... because for me, the 'cd' bind must obey what's in the HOME or HOMEPATH or whatever environment variable says where's the user's home directory. But now we are using what _Go thinks_ is the user's home...
What do you think?

internal/sh/shell_test.go
```diff
@@ -1360,75 +1442,69 @@ func TestExecuteRedirectionPipe(t *testing.T) {
...
-		return // Done
+	if msg := string(buf[:]); msg != message {
+		t.Fatalf("Unexpected message:\nGot:\t\t%s\nExpected:\t%s\n", msg, message)
 	}
 }
 
 func TestTCPRedirection(t *testing.T) {
-	testTCPRedirection(t, ":6666", `echo -n "hello world" >[1] "tcp://localhost:6666"`)
-	testTCPRedirection(t, ":6667", `echo -n "hello world" > "tcp://localhost:6667"`)
+	testTCPRedirection(t, ":4666", `echo -n "hello world" >[1] "tcp://localhost:4666"`)
```
tiago4orion: on windows, the non-privileged port range is 1024-5000 kkkk

nash_test.go
tiago4orion: future improvement =)

cmd/nash/cli.go
tiago4orion: yes, maybe verify it before calling setupCli(), something like:

```
if !noInit {
     err := setupCli(shell)
     ...
}
```

appveyor.yml
katcipis: =( =P

ast/node.go
katcipis: sorry I saw that later on the other MR =/

cmd/nash/cli.go
katcipis: Once I discussed a lot about this "one exit" vs "multiple exits", I prefer to return ASAP, it makes things easier for me to understand, but the guy argued that multiple returns are confusing (he defended only one always...which was bizarre for me). To be honest not sure if this is something related to how I think, but when I see a break depending on the size of the rest of the function I take more time to understand it than just returning.

internal/sh/shell.go
katcipis: It seems to me that what Go thinks is a platform independent home, no ? The question is more about being platform independent than coupling with a Go definition of home (does not seem like they are trying to define their own). At least it is what I get from the HomeDir docs.

cmd/nash/cli.go
tiago4orion: I wasn't talking about _pattern_ or the _one rule to rule them all_. I was just arguing for this specific case. No feelings against returning earlier as possible, I do this regularly, but don't think it's a rule.

internal/sh/shell.go
tiago4orion: yes! Thanks for the detailed review =)

internal/sh/shell.go
tiago4orion: Yes, but there's a subtle difference in the approach. Making cd obey the environment variable (instead of Go's way of getting this info) is that I can make something like:

```
λ> setenv HOME = "/"
λ> cd
λ> pwd
/
λ>
```

katcipis: I see, we could enforce to always be HOME, but scripts would have problem to be platform independent anyway I think, on windows it would need something like:

```
if is_windows {
    setenv HOME = $WHATEVER_WINDOWS_USES_AS_HOME
}
```

No idea what is better, but I see your point

tiago4orion:
:x: [Build nash 1.0.0.81 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.81) (commit https://github.com/NeowayLabs/nash/commit/b21bf80746 by @tiago4orion)

Issue #230, opened 2017-07-27
stdbin: have standard binaries distributed with the language


katcipis:
Small binaries (probably from enzo) that have more predictable behavior and nicer APIs

This way we can use commands on the stdlib and still have predictable behavior when executing the scripts on different systems.

We could use $NASHROOT/bin for this and search on it before searching on the host $PATH. This way the user do not have to change its own PATH and bash scripts continue to work without any impact (using the tools installed on the host).

tiago4orion:
awesome idea!

Will you address this in PR #228 also?

katcipis:
What about leaving it to another PR ? The stdlib one is already pretty big =)

tiago4orion:
:+1:

PR #240, opened 2017-09-27
stdbin: adding mkdir

tiago4orion:
This PR starts a long road to address #230. 

As I'm using a lot of windows nowadays... I'll start scratching my itches, by adding the programs I'm missing in windows. The work for the stdbin build system and integrated tests could come up in another PR.

Another PR could be created to copy our programs from c0defellas/enzo to here also.

p.s.: the name of the branch is `add-stdbin-pwd` because `pwd` was my initial target for this development, but soon realized I was unable to create the stdbin directory in windows without _right-click in explorer_ ... Then mkdir priority escalated very fast =P

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,9 @@ buildargs = -ldflags "-X main.VersionString=$(version)" -v
 all: build test install
 
 build:
-	go build $(buildargs) -o ./cmd/nash/nash ./cmd/nash
-	go build $(buildargs) -o ./cmd/nashfmt/nashfmt ./cmd/nashfmt
+	cd cmd/nash && go build $(buildargs) 
+	cd cmd/nashfmt && go build $(buildargs) 
+	cd stdbin/mkdir && go build $(buildargs)
 
 NASHPATH=$(HOME)/nash
 NASHROOT=$(HOME)/nashroot
@@ -23,6 +24,7 @@ install: build
 	cp -p ./cmd/nashfmt/nashfmt $(NASHROOT)/bin
 	rm -rf $(NASHROOT)/stdlib
 	cp -pr ./stdlib $(NASHROOT)/stdlib
+	cp -pr ./stdbin/mkdir/mkdir $(NASHROOT)/bin/mkdir
 
 deps:
 	go get -v -t golang.org/x/exp/ebnf
--- a/stdbin/mkdir/main.go
+++ b/stdbin/mkdir/main.go
@@ -22,7 +22,7 @@ func main() {
 	}
 	err := mkdirs(os.Args[1:])
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "error: %s\n", err.Error())
+		fmt.Fprintf(os.Stderr, "error: %s\n", err)
 		os.Exit(1)
 	}
 }
--- a/stdbin/mkdir/main.go
+++ b/stdbin/mkdir/main.go
@@ -1,0 +1,28 @@
+package main 
+
+import (
+	"os"
+	"fmt"
+)
+
+func mkdirs(dirnames []string) error {
+	for _, d := range dirnames {
+		if err := os.MkdirAll(d, 0644); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Fprintf(os.Stderr, "usage: %s <dir1> <dir2> ...\n", os.Args[0])
+		os.Exit(1)
+	}
+	err := mkdirs(os.Args[1:])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %s\n", err.Error())
+		os.Exit(1)
+	}
+}
--- a/stdbin/mkdir/mkdir_test.go
+++ b/stdbin/mkdir/mkdir_test.go
@@ -1,0 +1,40 @@
+package main 
+
+import (
+	"testing"
+	"path"
+	"path/filepath"
+	"io/ioutil"
+	"os"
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -7,8 +7,9 @@ buildargs = -ldflags "-X main.VersionString=$(version)" -v
 all: build test install
 
 build:
-	go build $(buildargs) -o ./cmd/nash/nash ./cmd/nash
-	go build $(buildargs) -o ./cmd/nashfmt/nashfmt ./cmd/nashfmt
+	cd cmd/nash && go build $(buildargs)
```
tiago4orion: this change was needed because on windows the generated binaries *must* have .exe extensions.

stdbin/mkdir/main.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Fprintf(os.Stderr, "usage: %s <dir1> <dir2> ...\n", os.Args[0])
+		os.Exit(1)
+	}
+	err := mkdirs(os.Args[1:])
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %s\n", err.Error())
```
katcipis: No need to call err.Error

stdbin/mkdir/main.go
```diff
@@ -0,0 +1,28 @@
+package main 
+
+import (
+	"os"
+	"fmt"
+)
+
+func mkdirs(dirnames []string) error {
+	for _, d := range dirnames {
+		if err := os.MkdirAll(d, 0644); err != nil {
```
katcipis: cant you pass os.ModeDir here instead of 0644 ?

katcipis: always creating all dirs seems nice, I always do it =).

another thing very common is to ignore if the dir already exists, in the end I always use "-p" for this. What you think about it being the default ?

stdbin/mkdir/mkdir_test.go
```diff
@@ -0,0 +1,38 @@
...
+		testdir := filepath.FromSlash(path.Join(tmpDir, p))
+		err = mkdirs([]string{testdir})
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if s, err := os.Stat(testdir); err != nil {
+			t.Fatal(err)
+		} else if s.Mode()&os.ModeDir != os.ModeDir {
+			t.Errorf("Invalid directory mode: %v", s.Mode())
```
katcipis: Why Errorf instead of Fatalf since all other checks results on Fatal ?

stdbin/mkdir/mkdir_test.go
```diff
@@ -0,0 +1,38 @@
...
+	testDirs := []string{
+		"1", "2", "3", "4", "5",
+		"some", "thing", "_random_",
+		"_",
+	}
+
+	defer os.RemoveAll(tmpDir)
+	for _, p := range testDirs {
+		testdir := filepath.FromSlash(path.Join(tmpDir, p))
+		err = mkdirs([]string{testdir})
```
katcipis: Isn't a good idea to test with arrays bigger than one too ? Also, what happens if the array is empty ? (can test it on the function or on the final cmd)

stdbin/mkdir/main.go
tiago4orion: yes, makes sense. That's what I commonly do also.

tiago4orion: > cant you pass os.ModeDir here instead of 0644
0644 are the permission bits here, not the file mode. Sorry, did I misunderstand you?

stdbin/mkdir/mkdir_test.go
tiago4orion: I'm idiot =)

stdbin/mkdir/mkdir_test.go
tiago4orion: yes, there are missing test cases. I made this in a rush... =/
thanks

stdbin/mkdir/main.go
katcipis: From the name I thought the os.ModeDir would be an integer with the permission bits (indicating the default permissions for directories). My mistake =)

stdbin/mkdir/mkdir_test.go
katcipis: Thanks for working on this =). We can add an TODO too ;-)

stdbin/mkdir/mkdir_test.go
katcipis: auheauheauheuaheauheauhe yeah like no one else does this kind of thing =P

Makefile
```diff
@@ -23,6 +24,7 @@ install: build
 	cp -p ./cmd/nashfmt/nashfmt $(NASHROOT)/bin
 	rm -rf $(NASHROOT)/stdlib
 	cp -pr ./stdlib $(NASHROOT)/stdlib
+	cp -pr ./stdbin/mkdir/mkdir $(NASHROOT)/bin/mkdir
```
tiago4orion: this install still only works on unix, I wll address this in the windows branch soon.

stdbin/mkdir/main.go
tiago4orion: 👍

Makefile
katcipis: thanks, I dont even have an env to test this kind of stuff =P (nor the patience)

tiago4orion:
Did you take a look at changes made?
I need this code in the `fix-windows` branch to finish the tests there :)

PR #242, opened 2017-09-28
Added 'rglob' function to stdlib.

vitorarins:


stdlib/rglob.sh
```diff
@@ -0,0 +1,29 @@
+fn dir_glob(dir, pattern, results) {
+	files <= find $dir -maxdepth 1 -mindepth 1
```
tiago4orion: stdlib functions should never abort the script. We must treat (or ignore in some cases) all errors.

tiago4orion: you can optimize this a lot passing `-type d` to get only the directories.

stdlib/rglob.sh
```diff
@@ -0,0 +1,29 @@
...
+	files <= split($files, "\n")
+
+	for f in $files {
+		_, status <= test -d $f
```
tiago4orion: if asking only directories to `find` then this check could be skipped. 
What do you think?

stdlib/rglob.sh
```diff
@@ -0,0 +1,29 @@
...
+			
+			results <= dir_glob($f, $pattern, $results)
+		}
+	}
+
+	return $results
+}
+
+fn rglob(pattern) {
+	working_dir <= pwd
```
tiago4orion: I dont think this is the right way to handle the rglob. It will only work in the current directory? I'm not saying we must add support to the '/path/**' syntax, but maybe the signature could be `rglob(dir, pattern)`. 
What do you guys think?

stdlib/rglob_test.sh
```diff
@@ -0,0 +1,85 @@
...
+	_, _      <= touch $test_txt
+	_, _      <= touch $test_sh
+
+	hello_dir = $temp_dir+"/hello"
+	world_dir = $temp_dir+"/world"
+
+	_, _      <= mkdir $hello_dir
+	_, _      <= mkdir $world_dir
+
+	honda_txt = $hello_dir+"/honda.txt"
```
tiago4orion: kkkkkkk

stdlib/rglob_test.sh
```diff
@@ -0,0 +1,85 @@
...
+	touch $test_txt
+	touch $test_sh
+
+	hello_dir = $temp_dir+"/hello"
+	world_dir = $temp_dir+"/world"
+
+	mkdir $hello_dir
+	mkdir $world_dir
+
+	honda_txt = $hello_dir+"/honda.txt"
```
tiago4orion: kkkkk

tiago4orion:
@vitorarins 
https://github.com/DoctorWkt/unix-jun72/blob/master/src/cmd/glob.c

Version 1 of unix had a program called `glob`, very useful =) Maybe we create a tool like that in stdbin to be used in conjunction with our `rglob` function.

katcipis:
@tiago4orion the idea is to remove the builtin glob function them ?

vitorarins:
> Maybe we create a tool like that in stdbin
Awesome! 

@katcipis @tiago4orion What do you think about creating a new glob that is capable of accepting the two asterisks (`**`) syntax?

stdlib/rglob_test.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env nash
+
+import rglob
+
+fn setup() {
+	temp_dir <= mktemp -d
```
tiago4orion: this directory is leaking.

stdlib/rglob_test.sh
```diff
@@ -0,0 +1,85 @@
...
+	honda_sh  = $hello_dir+"/honda.sh"
+	civic_txt = $world_dir+"/civic.txt"
+	civic_sh  = $world_dir+"/civic.sh"
+
+	touch $honda_txt
+	touch $honda_sh
+	touch $civic_txt
+	touch $civic_sh
+
+	return $temp_dir
```
tiago4orion: better to return the cleaning function also, something like:

```
       fn clean() {
	    rm -rf $temp_dir
        }

	return $temp_dir, $clean
```

tiago4orion:
@katcipis I don't think so. I prefer a builtin function that handles ** also, or a function in stdlib.. I sent the link to the unix glob because maybe we can create a helper binary to help the rglob nash function. But maybe this doesn't make sense.

katcipis:
@tiago4orion If the builtin function will continue to exist I see very little incentive to have a helper binary...not sure in what will exactly it help if we give support on the glob function.

PR #244, opened 2017-10-01
Add pwd

tiago4orion:
So simple... do makes sense testing this?
I made a test, but feels like I'm testing `os.Getwd` and `fmt.Println` from Go.. there's no much code here.
Makes sense?

Closes #244

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,9 @@ buildargs = -ldflags "-X main.VersionString=$(version)" -v
 all: build test install
 
 build:
-	go build $(buildargs) -o ./cmd/nash/nash ./cmd/nash
-	go build $(buildargs) -o ./cmd/nashfmt/nashfmt ./cmd/nashfmt
+	cd cmd/nash && go build $(buildargs) 
+	cd cmd/nashfmt && go build $(buildargs) 
+	cd stdbin/mkdir && go build $(buildargs)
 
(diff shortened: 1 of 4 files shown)


tiago4orion:
oops, the only code here is: https://github.com/NeowayLabs/nash/pull/244/files#diff-7e0bdb2b11a48277e00f1e8c58446148
the others are from #240 that I merged here also

Makefile
```diff
@@ -7,8 +7,9 @@ buildargs = -ldflags "-X main.VersionString=$(version)" -v
 all: build test install
 
 build:
-	go build $(buildargs) -o ./cmd/nash/nash ./cmd/nash
-	go build $(buildargs) -o ./cmd/nashfmt/nashfmt ./cmd/nashfmt
+	cd cmd/nash && go build $(buildargs) 
+	cd cmd/nashfmt && go build $(buildargs) 
+	cd stdbin/mkdir && go build $(buildargs)
```
vitorarins: This is going to be a big list :sweat_smile:

vitorarins:
@tiago4orion I agree that the code itself is really simple. But does it make sense to have a nash script testing the binary?

tiago4orion:
@vitorarins The code is obviously right. 

_'avoiding bugs by avoiding code' seems to be the only true fact of computer science_

PR #245, opened 2017-11-07
Add initial map implementation

katcipis:


tiago4orion:
tests are failing?

katcipis:
Not related to my changes =/

```
coverage: 77.5% of statements
ok  	github.com/NeowayLabs/nash/internal/sh	9.583s	coverage: 77.5% of statements
# github.com/NeowayLabs/nash/internal/sh/builtin_test
internal/sh/builtin/glob_test.go:84: Fatalf format %s has arg len(res) of wrong type int
FAIL	github.com/NeowayLabs/nash/internal/sh/builtin [build failed]
make: *** [test] Error 2
```

But I will fix it :D

katcipis:
Seems like go tip now has some extra validations for string formats... neat =)

katcipis:
The docs are missing, not sure if I document the map since I see it as an undesired hack =/.... perhaps later =P

tiago4orion:
I don't think docs are required right know also. Probably we will end with a syntatic map or a builtin one.

PR #246, opened 2017-11-08
Add indexing to strings

katcipis:


tests/internal/assert/equal.go
```diff
@@ -0,0 +1,11 @@
+// Package assert has come common assert functions
```
tiago4orion: typo

tests/internal/sh/shell.go
```diff
@@ -0,0 +1,48 @@
...
+	cmd := exec.Command(nashpath, scriptargs...)
+
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err = cmd.Run()
+	output := stdout.String() + stderr.String()
```
vitorarins: I am not sure concatenating these is the best approach for validating output and error. I can't see any use case right now so I don't think it's critical to change this now.

sh/obj.go
```diff
@@ -29,25 +29,88 @@ type (
...
 		objType
-		str string
+		runes []rune
+	}
+
+	Enumerable interface {
+		Len() int
+	}
+
+	WriteIndexer interface {
```
tiago4orion: I don't think Write/Read are good names in this case, to specify the kind of indexer. I prefer SetIndexer and GetIndexer. Or, if the interface has only a Get and a Set, then just Getter and Setter, respectively.

sh/obj.go
```diff
@@ -29,25 +29,88 @@ type (
...
 	}
 }
 
-func (o *StrObj) Str() string { return o.str }
+func (o *StrObj) Str() string { return string(o.runes) }
 
 func (o *StrObj) String() string { return o.Str() }
 
+func (o *StrObj) Get(index int) (Obj, error) {
+	// FIXME: Use runes instead
```
tiago4orion: fixed, no?

katcipis: yeah, I forgot =P

sh/obj.go
katcipis: I thought about that but it seemed too javanesque for your taste...it is odd that you are asking for these names =P

katcipis: @vitorarins what you think ? These ones are really hard to name x_x

tiago4orion: haha No man, I have no problem with that. The meaning is clear, they're interface to *set* and *get* things.

vitorarins: I think Java should burn in hell :fire: just kidding (or am I?)

Set and GetIndexer seems more intuitive for me. But I don't have a problem with the current name.

katcipis: What about:

```go
	Collection interface {
		Len() int
	}

	WriteableCollection interface {
                Collection
		Set(index int, val Obj) error
	}

	ReadableCollection interface {
		Collection
		Get(index int) (Obj, error)
	}
```

or

```go
	Collection interface {
		Len() int
	}

	CollectionWriter interface {
                Collection
		Set(index int, val Obj) error
	}

	CollectionReader interface {
		Collection
		Get(index int) (Obj, error)
	}
```

or

```go
	Collection interface {
		Len() int
	}

	CollectionSetter interface {
                Collection
		Set(index int, val Obj) error
	}

	CollectionGetter interface {
		Collection
		Get(index int) (Obj, error)
	}
```

or

```go
	Collection interface {
                Get(index int) (Obj, error)
		Len() int
	}

	WriteableCollection interface {
                Collection
		Set(index int, val Obj) error
	}
```

It seems like the minimum set of operations on a collection is at least Len + Get, so the last one has some appealing to me.

katcipis: @tiago4orion @vitorarins take a look =)

vitorarins: The last one seems great!

tiago4orion: something *writeable* that not resembles an io.Writer is weird for me. I prefer Collection/CollectionGetter/CollectionSetter.

katcipis: I agree with @vitorarins on the last one.

PR #248, opened 2017-11-12
WIP: Add formal specification of the language

tiago4orion:
This PR introduces the formal specification of the language.
What do you guys think? Is it better than EBNF?
The link to ohm editor with nash's gramar is here:
https://ohmlang.github.io/editor/#fca725147857c09bb442be1ce03048e0

Yes, there's some js to validate the grammar until ohm has a Go parser.
Still thinking the best way to install ohm-js dependency and run the grammar check from 'make test'.
Maybe ran with docker?

### DIFF
--- a/spec/assignments.sh
+++ b/spec/assignments.sh
@@ -1,0 +1,49 @@
+
+# Single assignment
+var a = "";
+var a = "something";
+var a = "/usr/home/ken";
+var a = "C:\\Users\\Bill";
+var interests = (
+    "plan9"
+    "go"
+    "c"
+    "asm"
+    "scheme"
+);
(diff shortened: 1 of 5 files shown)


ryukinix:
Really interesting thing! I miss a lot checkers of EBNF grammar, I search a lot and didn't find anything really good; mostly just legacy, dead or just too painful too use. This ohm is a quite nice. I'm still think that ebnf is nice to read as human being, but have all that software of ohm doing the parsing, show-case of AST and other stuff is really great!

Seems pretty fine to me running the ohm tests inside of docker through make test. Formal tests it's important.

PR #256, opened 2018-01-13
add common cfg into tests pkg and simplify setup of tests

### LINKED ISSUE #255: go >= 1.8 doesnt requires GOPATH
new machine, then got this:

```
λ> make
cd cmd/nash && go build -ldflags "-X main.VersionString=1c720002ec961bb61275722deb010c9403d3f935" -v 
github.com/NeowayLabs/nash/token
github.com/NeowayLabs/nash/sh
github.com/NeowayLabs/nash/cmd/nash/vendor/github.com/chzyer/readline
github.com/NeowayLabs/nash/scanner
github.com/NeowayLabs/nash/ast
github.com/NeowayLabs/nash/errors
github.com/NeowayLabs/nash/internal/sh/builtin
github.com/NeowayLabs/nash/parser
github.com/NeowayLabs/nash/internal/sh
github.com/NeowayLabs/nash
github.com/NeowayLabs/nash/cmd/nash
cd cmd/nashfmt && go build -ldflags "-X main.VersionString=1c720002ec961bb61275722deb010c9403d3f935" -v 
github.com/NeowayLabs/nash/token
github.com/NeowayLabs/nash/scanner
github.com/NeowayLabs/nash/ast
github.com/NeowayLabs/nash/errors
github.com/NeowayLabs/nash/parser
github.com/NeowayLabs/nash/cmd/nashfmt
cd stdbin/mkdir && go build -ldflags "-X main.VersionString=1c720002ec961bb61275722deb010c9403d3f935" -v
github.com/NeowayLabs/nash/stdbin/mkdir
go get -v -t golang.org/x/exp/ebnf
Fetching https://golang.org/x/exp/ebnf?go-get=1
Parsing meta tags from https://golang.org/x/exp/ebnf?go-get=1 (status code 200)
get "golang.org/x/exp/ebnf": found meta tag get.metaImport{Prefix:"golang.org/x/exp", VCS:"git", RepoRoot:"https://go.googlesource.com/exp"} at https://golang.org/x/exp/ebnf?go-get=1
get "golang.org/x/exp/ebnf": verifying non-authoritative meta tag
Fetching https://golang.org/x/exp?go-get=1
Parsing meta tags from https://golang.org/x/exp?go-get=1 (status code 200)
golang.org/x/exp (download)
golang.org/x/exp/ebnf
./hack/check.sh
panic: Please, run tests from inside GOPATH

goroutine 1 [running]:
github.com/NeowayLabs/nash.init.0()
	/home/i4k/go/src/github.com/NeowayLabs/nash/nash_test.go:22 +0x28c
github.com/NeowayLabs/nash.init()
	<autogenerated>:1 +0xd2
main.init()
	<autogenerated>:1 +0xaf
exit status 2
FAIL	github.com/NeowayLabs/nash	0.006s
Makefile:42: recipe for target 'test' failed
make: *** [test] Error 1
ERROR: exit status 2
```

tiago4orion:
@katcipis I know the tests package are there for blackbox tests, but I think it's a good place to put some of the common/util logics that were currently repeated in test code (like getting gopath, testdir, nashcmd).

What do you think?

Closes #255

### DIFF
--- a/nash_test.go
+++ b/nash_test.go
@@ -6,32 +6,14 @@ import (
 	"testing"
 
 	"github.com/NeowayLabs/nash/sh"
+	"github.com/NeowayLabs/nash/tests"
 )
 
 // only testing the public API
 // bypass to internal sh.Shell
 
-var (
-	gopath, testDir, nashdPath string
-)
-
-func init() {
-	gopath = os.Getenv("GOPATH")
-
-	if gopath == "" {
-		panic("Please, run tests from inside GOPATH")
-	}
-
-	testDir = gopath + "/src/github.com/NeowayLabs/nash/" + "testfiles"
-	nashdPath = gopath + "/src/github.com/NeowayLabs/nash/cmd/nash/nash"
-
-	if _, err := os.Stat(nashdPath); err != nil {
-		panic("Please, run make build before running tests")
-	}
-}
-
 func TestExecuteFile(t *testing.T) {
-	testfile := testDir + "/ex1.sh"
+	testfile := tests.Testdir + "/ex1.sh"
 
 	var out bytes.Buffer
 
@@ -42,7 +24,7 @@ func TestExecuteFile(t *testing.T) {
 		return
 	}
 
-	shell.SetNashdPath(nashdPath)
+	shell.SetNashdPath(tests.Nashcmd)
 	shell.SetStdout(&out)
 	shell.SetStderr(os.Stderr)
 	shell.SetStdin(os.Stdin)
--- a/tests/cfg.go
+++ b/tests/cfg.go
@@ -1,3 +1,62 @@
 package tests
 
-const nashcmd = "../cmd/nash/nash"
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+)
+
+var (
+	// Nashcmd is the nash's absolute binary path in source
+	Nashcmd string
+	// Testdir is the test assets directory
+	Testdir string
+	Gopath  string
+)
+
+func getGopath() (string, error) {
+	gopathenv := os.Getenv("GOPATH")
+	if gopathenv != "" {
+		return gopathenv, nil
+	}
+
+	usr, err := user.Current()
+	if err != nil {
+		return "", fmt.Errorf("failed to get user's home directory: %s", err)
+	}
+
+	gopathhome := filepath.Join(usr.HomeDir, "go")
+	if _, err := os.Stat(gopathhome); err != nil {
+		return "", errors.New("gopath not found")
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("failed to get working directory: %s", err)
+	}
+
+	if !strings.HasPrefix(wd, gopathhome) {
+		return "", errors.New("Run tests require code inside gopath")
+	}
+	return gopathhome, nil
+}
+
+func init() {
+	gopath, err := getGopath()
+	if err != nil {
+		panic(err)
+	}
+
+	Gopath = gopath
+	Testdir = filepath.Join(Gopath, "src", "github.com",
+		"NeowayLabs", "nash", "testfiles")
+	Nashcmd = filepath.Join(Gopath, "src", "github.com",
+		"NeowayLabs", "nash", "cmd", "nash", "nash")
+
+	if _, err := os.Stat(Nashcmd); err != nil {
+		panic("Please, run make build before running tests")
+	}
+}
--- a/internal/sh/shell_test.go
+++ b/internal/sh/shell_test.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/NeowayLabs/nash/sh"
+	"github.com/NeowayLabs/nash/tests"
 )
 
 type execTestCase struct {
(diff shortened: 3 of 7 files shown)


nash_test.go
```diff
@@ -42,7 +24,7 @@ func TestExecuteFile(t *testing.T) {
 		return
 	}
 
-	shell.SetNashdPath(nashdPath)
+	shell.SetNashdPath(tests.Nashcmd)
```
katcipis: Perhaps having a function that creates a shell already with the nashd path set is even more useful

tests/cfg.go
```diff
@@ -1,3 +1,62 @@
...
+	// Nashcmd is the nash's absolute binary path in source
+	Nashcmd string
+	// Testdir is the test assets directory
+	Testdir string
+	Gopath  string
+)
+
+func getGopath() (string, error) {
+	gopathenv := os.Getenv("GOPATH")
+	if gopathenv != "" {
```
katcipis: When there is a GOPATH set we are not going to check if the code is inside the GOPATH ?

tests/cfg.go
```diff
@@ -1,3 +1,62 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get user's home directory: %s", err)
+	}
+
+	gopathhome := filepath.Join(usr.HomeDir, "go")
+	if _, err := os.Stat(gopathhome); err != nil {
+		return "", errors.New("gopath not found")
+	}
+
+	wd, err := os.Getwd()
```
katcipis: Here we validate if the code is inside. Perhaps we could extract this in a function and use on all cases (GOPATH and user home GOPATH). Or is there a reason to check this only when the user derived GOPATH is used ?

tests/cfg.go
```diff
@@ -1,3 +1,62 @@
...
+	if !strings.HasPrefix(wd, gopathhome) {
+		return "", errors.New("Run tests require code inside gopath")
+	}
+	return gopathhome, nil
+}
+
+func init() {
+	gopath, err := getGopath()
+	if err != nil {
+		panic(err)
```
katcipis: I know it is more pratical to use, but I really dislike packages that can panic just by the action of importing them =/. The solution to this would be to have some sort of NewCfg function that creates new configurations, the callers will have more trouble using the cfg but we won't have the surprise of code that panics before even executing. In this context of tests perhaps it is ok (on a prod code I imagine the main function not even entering, for me it is a little confusing).

katcipis: Also creating configurations explicitely enforces isolation between tests, since the variables can't be constants they can be changed at runtime.

katcipis:
I think it is a great idea =D, just made some remarks on some of the implementation details. It would be good to hear what @vitorarins thinks too =)

tests/cfg.go
vitorarins: I agree that is ok for this context of testing. But it would also be nice to have this NewCfg function.

tests/cfg.go
tiago4orion: yes, I'm idiot

tests/cfg.go
tiago4orion: But there's nothing to configure here. There's only environment's detection and aborts if something is not expected. Or maybe I didn't understand what you guys are meaning.

Everyone that imports "test" would have this environment's information for free, don't require instantiation of NewCfg at every package tests.

About panic'ing on init, obviously, this is ok because it's testing. I hate packages that panics in production.

katcipis: You are not configuring something, you are loading a configuration. An example is where is the nash binary to be used on the tests, it does not matter if it is a command line argument or an environment variable, it seems like a configuration to me (not sure what would I call it if not a configuration).The whole logic of loading env vars and evaluating if we have a GOPATH or need to infer one etc is the action of loading a configuration for me.

About everyone having the information for free this is really true, it is easier to use, but also easier for one test to interfere with another one and even make test ordering affect results since it is shared mutable state (the choice would be easier if they where constants). But in this case it does not seem to be terrible. The tradeoff reminds me of Limbo's module system, improving isolation almost always makes usability a little harder =P

tiago4orion: Configuration is something you can change to get different results. The kind of data collected here is like 'info' (for me), if you change it for something else you break the build/test.

vitorarins: I think it would be nice to have something more dynamic in the future. Right now I agree with @tiago4orion that there isn't much to configure yet.

katcipis: I can't change my GOPATH ? since we are loading this (or assuming one from user's home if absent) for me it is a configuration loading =)

katcipis: I'm not insisting that we should extract the config in a function, just that I see that the act of loading environment variables as loading configuration. Not sure it will be always that, but usually is and in the case of GOPATH it really seems to be a config. Differences in GOPATH may use different versions of dependencies if you have two different Go src folders, this may not affect nash tests but qualifies it as a configuration for me.

tiago4orion: I think you're not able to configure GOPATH at this point either. Here we are only making sure you are building the software inside GOPATH. We make tests.Gopath available to avoid detect it every time (because of fallback to ~/go if not set). We don't matter this value, only that code is inside it because we build absolute paths based on it (the testing directory and Nash cmd path).

But, this works making it a configuration also but don't see why at this point.

PR #264, opened 2018-03-30
print init loading fatal error instead of exiting

katcipis:
A lot of times I shot myself on the foot being unable to start terminals on taowm and rio and not knowing why they where not loading (could be a lot of different things) but was actually some error loading my nash init.sh. It seems better to drop the user on a default prompt so he can fix any kind of mistake instead of just exiting the process (which causes the entire terminal to close).

### DIFF
--- a/cmd/nash/cli.go
+++ b/cmd/nash/cli.go
@@ -65,9 +65,8 @@ func importInitFile(shell *nash.Shell, initFile string) (bool, error) {
 	return false, nil
 }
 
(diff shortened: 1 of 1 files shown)


tiago4orion:
:x: [Build nash 1.0.0.165 failed](https://ci.appveyor.com/project/tiago4orion/nash/build/1.0.0.165) (commit https://github.com/NeowayLabs/nash/commit/407848b804 by @katcipis)

PR #268, opened 2018-06-26
Fix release generation for all OSs

katcipis:
I'm not sure if other OSes will have /dev/stdout as linux, but it seemed that the outlier is windows =P. So I tried to fix the write stdbin on these other platforms. I also added the release procedure to the CI to avoid this is the future =)

### DIFF
--- a/hack/releaser.sh
+++ b/hack/releaser.sh
@@ -5,9 +5,9 @@ if len($ARGS) != "2" {
 	exit("1")
 }
 
-version        = $ARGS[1]
-supported_os   = ("linux" "darwin" "windows")
-supported_arch = ("amd64")
+var version        = $ARGS[1]
+var supported_os   = ("linux" "darwin" "windows")
+var supported_arch = ("amd64")
 
 # Guarantee passing tests at least on the host arch/os
 make test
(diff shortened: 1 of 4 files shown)


stdbin/write/fd_windows.go
```diff
@@ -5,9 +5,9 @@ import (
 	"os"
 )
 
-func specialFile(path string) (io.Writer, bool) {
-	if fname == "CON" { // holycrap!
```
tiago4orion: o.o
err... this is being used?

hack/releaser.sh
```diff
@@ -48,36 +49,37 @@ for os in $supported_os {
...
-		projectdir <= pwd
-		distar  <= format("%s/dist/nash-%s-%s-%s.tar.gz", $projectdir, $version, $os, $arch)
+		var projectdir <= pwd
+		var distar  <= format("%s/dist/nash-%s-%s-%s.tar.gz", $projectdir, $version, $os, $arch)
 
 		chdir($pkgdir)
-		pkgraw <= ls
-		pkgfiles <= split($pkgraw, "\n")
+		var pkgraw <= ls
+		var pkgfiles <= split($pkgraw, "\n")
```
tiago4orion: lots and lots and lots of var =P

katcipis: yeah =/

stdbin/write/fd_windows.go
katcipis: I dont know...but the code was there and the build was failing =P. I tried to fix it, but I have nothing against just removing it. Your call =P

PR #269, opened 2018-06-28
Dont abort on invalid NASHPATH/NASHROOT

katcipis:
Instead of aborting just print error messages on the shell stderr so the user dont go completely blind without an stdlib.

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -153,7 +153,11 @@ func NewShell(
 	
 	err = validateDirs(nashpath, nashroot)
 	if err != nil {
-		shell.Stderr().Write([]byte(err.Error() + "\n"))
+		printerr := func(msg string) {
+			shell.Stderr().Write([]byte(msg + "\n"))
+		}
+		printerr(err.Error())
+		printerr("please check your NASHPATH and NASHROOT so they point to valid locations")
 	}
 	
 	return shell, nil
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -62,7 +62,7 @@ type (
 		parent *Shell
 
 		repr string // string representation
-		
+
 		nashpath string
 		nashroot string
 
@@ -140,8 +140,8 @@ func NewShell(
 		Mutex:       &sync.Mutex{},
 		sigs:        make(chan os.Signal, 1),
 		filename:    "<interactive>",
-		nashpath: nashpath,
-		nashroot: nashroot,
+		nashpath:    nashpath,
+		nashroot:    nashroot,
 	}
 
 	err := shell.setup()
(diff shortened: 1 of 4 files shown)


internal/sh/shell.go
```diff
@@ -117,22 +117,23 @@ func (e *errStopWalking) StopWalking() bool { return true }
 // NewShell creates a new shell object
 // nashpath will be used to search libraries and nashroot will be used to
 // search for the standard library shipped with the language.
-func NewShell(nashpath string, nashroot string) (*Shell, error) {
+func NewShell(
```
tiago4orion: 90% of times, stdin, stdout and stderr points to the standard locations. The refactor of `NewShell` usage reflect this.

internal/sh/shell.go
```diff
@@ -2450,15 +2461,15 @@ func (shell *Shell) executeIf(n *ast.IfNode) ([]sh.Obj, error) {
 
 func validateDirs(nashpath string, nashroot string) error {
 	if nashpath == nashroot {
-		return fmt.Errorf("invalid nashpath[%s] == nashroot[%s], they must differ", nashpath, nashroot)
+		return fmt.Errorf("invalid nashpath and nashroot, they are both[%s] but they must differ", nashpath)
 	}
 	err := validateDir(nashpath)
 	if err != nil {
-		return fmt.Errorf("error[%s] validating nashpath", err)
+		return fmt.Errorf("error[%s]: invalid nashpath, history won't be saved", err)
```
tiago4orion: I think a better message is: 
_Invalid NASHPATH, user's config won't be loaded_

About the history file, probably needs to disable the history in the cli.go also, because otherwise the readline lib could crash. Have you tested what happens to the history when nashpath is not set?

katcipis: I assume the history won't exist =P. I used for a while and it works just fine, since the path I used was completely invalid I suppose no history is being taken and I didn't see any adverse effect on usage

katcipis:
We could improve later, but I think this solves some problems that @lborguetti had.

katcipis:
@tiago4orion anything against merging this ?

tiago4orion:
Now in the internal Shell, the *Standard* file descriptors of each shell could be customized in the constructor. I was waiting for your feelings regarding this change.

For now, a subshell imports the standard descriptors from the parent shell, then the only place that requires custom std I/O are the tests.
Why not keep the constructor args small and create a new custom constructor for that?

```go
NewShell(nashpath, nashroot string) (*Shell, error)
NewCustomShell(nashpath, nashroot string, in io.Reader, out, err io.Writer) (*Shell, error)
```

That way, only changes to the internal tests would be required, no?
I don't like when an API requires that optionals (things that probably are always the same) are required upfront.
But these are not *strong feelings* as this is an internal API. I'm ok to merge if you think the change makes sense.

katcipis:
I was not very happy with these tests anyway, left to think about a better way to check the import errors later x_x

PR #271, opened 2018-11-12
Fix Releaser script

vitorarins:
- Add CGO_ENABLED depending on the os; 
- Nashfmt'd the script (sorry).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ sudo: false
 
 go:
   - "tip"
-  - "1.10"
-  - "1.9"
+  - "1.12"
 
 install:
   - go get -v golang.org/x/exp/ebnf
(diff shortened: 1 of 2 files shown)


katcipis:
It worked fine here with Go 1.2 ...that is odd

katcipis:
@vitorarins specially since it happened on the tip perhaps it was some instability...or some non-determinism on the test. I changed the CI to use only Go >= 1.12 (using some Go 1.12 stdlib stuff on other branches) and it worked fine =).

vitorarins:
I've been removed from neowaylabs organization on Docker hub, so I can't check why the image is failing to build on docker cloud.
But I think we can merge this.

katcipis:
Why the fuck  would we remove you ? ¬¬

You have been re-added. I just checked that the Go version in the image is pretty old (1.8)...it could be something that we are using that is newer than that. I think even better type inference is newer than 1.8.

vitorarins:
Ok thanks :)

So actually the Dockerfile is based on the old build process.
We need to update it to use the install script or something.

PR #274, opened 2019-03-02
Use Go 1.12 new function to get status code of processes

katcipis:


internal/sh/builtin/exit_test.go
```diff
@@ -44,25 +38,14 @@ func TestExit(t *testing.T) {
...
-					got := status.ExitStatus()
-					if desc.result != got {
-						t.Fatalf("expected[%d] got[%d]", desc.result, got)
-					}
-				} else {
-					t.Fatal("exit result does not have a  ExitStatus method")
-				}
-			} else {
-				t.Fatalf("unexpected error: %v", err)
+			got := cmd.ProcessState.ExitCode()
```
tiago4orion: Much better now =) I wonder if they use this same approach now behind the scenes...

tiago4orion:
@katcipis Could you update the Dockerfile to use go1.12 also?
https://cloud.docker.com/u/neowaylabs/repository/docker/neowaylabs/nash/builds/71824bd8-1657-41d5-91a4-27c5bdf96057

internal/sh/builtin/exit_test.go
katcipis: Good question, will take a look =)

PR #279, opened 2019-06-16
OSX symlinks for /tmp and /var break the tests

tiago4orion:
On OSX, the directories /tmp and /var are symlinks to
/private/tmp and /private/var respectively.
This change adds an additional call to filepath.EvalSymlinks
before using paths obtained by TempDir.

Also, this change identified a problem related to logging of
the warnings when nashpath and/or nashroot aren't properly set.
As there are tests that do instantiate nash with wrong directories
for testing, this was making a mess in the stderr of the tests
because now nash doesn't abort in case of errors. This change address that too. (Maybe not in the best way)

Signed-off-by: Tiago de Bem <tiago.moura@nginx.com>

### DIFF
--- a/internal/sh/shell.go
+++ b/internal/sh/shell.go
@@ -41,6 +41,7 @@ type (
 		name        string
 		debug       bool
 		interactive bool
+		abortOnErr  bool
 		logf        LogFn
 		nashdPath   string
 		isFn        bool
@@ -114,14 +115,22 @@ func newErrStopWalking() *errStopWalking {
 
 func (e *errStopWalking) StopWalking() bool { return true }
 
+func NewAbortShell(nashpath string, nashroot string) (*Shell, error) {
+	return newShell(nashpath, nashroot, true)
+}
+
 // NewShell creates a new shell object
 // nashpath will be used to search libraries and nashroot will be used to
 // search for the standard library shipped with the language.
 func NewShell(nashpath string, nashroot string) (*Shell, error) {
+	return newShell(nashpath, nashroot, false)
+}
 
+func newShell(nashpath string, nashroot string, abort bool) (*Shell, error) {
 	shell := &Shell{
 		name:        "parent scope",
 		interactive: false,
+		abortOnErr:  abort,
 		isFn:        false,
 		logf:        NewLog(logNS, false),
 		nashdPath:   nashdAutoDiscover(),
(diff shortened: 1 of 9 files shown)


tiago4orion:
@vitorarins As we don't have CI for OSX, could you execute the tests on your laptop?

internal/sh/shell.go
```diff
@@ -144,9 +153,12 @@ func NewShell(nashpath string, nashroot string) (*Shell, error) {
 	}
 
 	shell.setupSignals()
-
 	err = validateDirs(nashpath, nashroot)
 	if err != nil {
+		if shell.abortOnErr {
```
katcipis: Man I think I made really bad choices in the past when dealing with this and now the bad choices are generating more bad choices (as bad choices usually do, nasty things).

I think we are suffering because we are not modelling errors properly. Even some tests could be improved to check for error behavior/type instead of the contents of the error as a string. But in this particular case, I was thinking on always returning an error. But when the error is recoverable, in the sense that nash can continue execution after it, it could return a "recoverable error". The details on how the error is built could be encapsulated in the package and the caller would have to just do this:

```go
arinsLindo, err := sh.NewShell()
if err != nil {
    if !sh.IsRecoverableErr(err) {
        // Abort here
    }
    // Just log the error for the user and goes on
}
```

In this case the behavior of the interactive CLI is guaranteed but the printing of the error is not embedded inside the sh package, which is awful x_x. And in the tests we can check for any kind of error because we dont want to differentiate between recoverable/unrecoverable errors.

Perhaps we could start our own error package, it would be like:

```go
arinsLindo, err := sh.NewShell()
if err != nil {
    if !errors.IsRecoverableErr(err) {
        // Abort here
    }
    // Just log the error for the user and goes on
}
```

And we could evolve to test things like the operation that generated an error, etc.

Something like https://commandcenter.blogspot.com/2017/12/error-handling-in-upspin.html

katcipis: We could also just work with error types and let the user decide which error types are fatal/recoverable or not. But it seems to me that the sh package will be able to define what is recoverable better than the caller (it knows how it works and what it can recover for etc).

tiago4orion: I agree with you. But just wanted to solve the problem without having to start a discussion of design. I like your idea.

tiago4orion: I could implement in an upcoming PR

PR #282, opened 2020-07-21
Update Tiago Natel email

katcipis:
I was taking a look on the Travis files...and now it actually seems that we don't have CI anymore, did we remove it ? (don't remember that). We probably lost it when we moved the projects across groups...I think :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,5 +32,5 @@ notifications:
     on_failure: always
     on_start: never
   email:
-    - tiago.natel@neoway.com.br
+    - tiago4orion@gmail.com
     - tiagokatcipis@gmail.com


katcipis:
Actually the builds are still happening, I think it is just the github PR integration that is broken: https://travis-ci.org/github/madlambda/nash/builds/710419331

vitorarins:
> Actually the builds are still happening, I think it is just the github PR integration that is broken: https://travis-ci.org/github/madlambda/nash/builds/710419331

[This comment](https://github.com/travis-ci/travis-ci/issues/6602#issuecomment-247029548) suggests we need to do a "sync" for the old and new repo. But it's from 2016 :confused:

PR #265, opened 2018-05-03
Refactor nashpath and nashroot

katcipis:
Extract env var loading logic out of the shell instance and load it only on main

PR #283, opened 2020-07-26
Add module support

katcipis:
To proper support modules I removed the readlines from the vendor since they are a modded/forked version of it, not just a vendor. Also needed to fix project root detection. Given how modules are implemented I think it is reasonably safe to use the go.mod file as an anchor to finding the root (you can't support modules without it).

Just cloning the project on any place in your file system, without a GOPATH defined, and running **make install** should work now.


katcipis:
And it seems... that by magic travis integration is working again o.O

tests/cfg.go
```diff
@@ -40,6 +33,58 @@ func init() {
...
+	// We used to use GOPATH as a way to infer the root of the
+	// project, now with Go modules this doesn't work anymore.
+	// Since module definition files only appear on the root
+	// of the project we use them instead, recursively going
+	// backwards in the file system until we find them.
+	//
+	// From: https://blog.golang.org/using-go-modules
+	// A module is a collection of Go packages stored in a file tree with a go.mod file at its root
+	//
+	// RIP GOPATH :-(
```
i4ki, line 52: 😞

vendor/golang.org/x/exp/PATENTS
```diff
@@ -0,0 +1,22 @@
+Additional IP Rights Grant (Patents)
```
i4ki, line 1: 🤔

vendor/github.com/chzyer/test/test.go
```diff
@@ -0,0 +1,420 @@
...
+		return int64(n), true
+	case int32:
+		return int64(n), true
+	case int64:
+		return int64(n), true
+	case uintptr:
+		return int64(n), true
+	default:
+		return -1, false
+	}
```
i4ki, line 210:
```
func toInt(a interface{}) (int64, bool) {
    switch n := a.(type) {
    case int, int8, int16, int32, int64, uintptr:
        return int64(n), true
    default:
        return -1, false
    }
}
```
🤣

tests/cfg.go
```diff
@@ -40,6 +33,58 @@ func init() {
...
+
+	dir, err := os.Getwd()
+	separator := string(filepath.Separator)
+	if err != nil {
```
vitorarins, line 56: Sorry if this is a stupid question, but why not check for the error immediately after `Getwd()`?

katcipis: Because I'm stupid :-) (and your question is not, thanks :heart:

katcipis:
> Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.

I just have a vague memory about this, did we really drop support for Windows @i4k ?

vendor/golang.org/x/exp/PATENTS
katcipis: :thinking: :thinking:

tests/cfg.go
katcipis: It took me almost an year to accept that :-(. At least vendoring still works...kinda =P

i4ki:
> > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> 
> I just have a vague memory about this, did we really drop support for Windows @i4k ?

this is news to me but I don't care anymore.

katcipis:
> > > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> > 
> > 
> > I just have a vague memory about this, did we really drop support for Windows @i4k ?
> 
> this is news to me but I don't care anymore.

How careless =P

i4ki:
> > > > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> > > 
> > > 
> > > I just have a vague memory about this, did we really drop support for Windows @i4k ?
> > 
> > 
> > this is news to me but I don't care anymore.
> 
> How careless =P

you tagged the wrong i4k. ;-)

katcipis:
> > > > > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> > > > 
> > > > 
> > > > I just have a vague memory about this, did we really drop support for Windows @i4k ?
> > > 
> > > 
> > > this is news to me but I don't care anymore.
> > 
> > 
> > How careless =P
> 
> you tagged the wrong i4k. ;-)

:rofl:

vitorarins:
> > > > > > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> > > > > 
> > > > > 
> > > > > I just have a vague memory about this, did we really drop support for Windows @i4k ?
> > > > 
> > > > 
> > > > this is news to me but I don't care anymore.
> > > 
> > > 
> > > How careless =P
> > 
> > 
> > you tagged the wrong i4k. ;-)

:rofl:

vitorarins:
> > > Looks like there is a lot of Windows stuff in readline we can get rid of in the future, since we're not supporting Windows anymore.
> > 
> > 
> > I just have a vague memory about this, did we really drop support for Windows @i4k ?
> 
> this is news to me but I don't care anymore.

Well, I also remember vaguely about the discussion we had offline about this.
But looking at the docs there is a section about environments where nash is tested, and those are all Linux.
Also we dropped OSX support in this PR #265 So for a moment I thought idea was to just focus on Linux

But anyway, was just an idea. Maybe we can bring back a proper support for Windows and OSX.

i4ki:
We can drop windows for sure, but we need to bring back OSX.

katcipis:
I think the release script still builds the binary for all platforms: 

https://github.com/madlambda/nash/blob/master/hack/releaser.sh#L9

Also the latest release has darwin/win:

https://github.com/madlambda/nash/releases/tag/v1.1

